PR #1, opened 2017-09-22
Auto stash before merge of "develop" and "origin/develop"

koyan:
Initial commit of the block for happy point. Note: The css from the template file will have to be removed after we have the master css in place. The template file still needs work for the overides.


27Bslash6:
May I ask why isn't twig being used for presentation and logic?  Defeats the purpose of having the extra layer if it's not going to be used for exactly this purpose.

koyan:
As we said in the dev channel, we want the shortcode plugin blocks to be able to be used independently of the specific template. So, after discussing it with the whole team (and feedback from Adrian) we decided that the front end parts of the plugin should not be in twig.

27Bslash6:
If you are planning to re-use these blocks for anything vaguely related to planet4, I think it safe to assume we will have the Timber plugin installed.  While we might postulate future reuse of these plugins in non-planet4 Wordpress installations, I think it not unreasonable to also expect that we use Timber and the twig syntax uniformly across all Wordpress installs going forwards.  Technically, it's a trivial element to include in the infrastructure, and well worth maintenance due to its functionality.  

Further, if these modules are to be for generic re-use in non-Planet-4, then might we first need to change the name of the repository?  Or is Timber being reconsidered project-wide?  I'm of the opinion that if you're going to use Timber in one element of the site, then make it a standard - all markup happens via Twig syntax using the Timber render method.  Proper separation of application logic and presentation.

I'm afraid I just don't see any real benefit of dropping Timber for such core elements as will be included in this repository, as it is such a useful tool for:
- Separation of logic and presentation
- Modular, reusable presentation blocks
- Clear and readable presentation syntax
- Auto-escaped variables and other helper functionality
- Modular configurable object, transient, compiled template caching

 What, particularly at this point of the project, is being gained by not using it?

koyan:
After a long discussion with several members of the team , I found where this was discussed, what was said, and at the end it was never said not to use twig (It was a different discussion). Thank you for pointing it out Ray. I will change the code and push a new version of the template (using twig)

includes/blocks/happy_point.twig
```diff
@@ -0,0 +1,37 @@
...
+						{{ fields['background'] }}
+					{% endif %}
+					{% if ( fields['boxout_title'] ) %}
+						<h3>{{ fields['boxout_title']|e('html') }}</h3>
+					{% endif %}
+					{% if ( fields['boxout_descr'] ) %}
+						<h4>{{ fields['boxout_descr']|e('html') }}</h4>
+					{% endif %}
+					{% if (( fields['boxout_link_url'] ) and (fields['boxout_link_text'])) %}
+						<a href="<?php echo $fields['boxout_link_url']; ?>" class="btn btn-primary"><?php echo $fields['boxout_link_text']; ?></a>
```
angtheod: probably forgot to change this with

<a href="{{ fields['boxout_link_url']|e('html_attr') }}" class="btn btn-primary">{{ fields['boxout_link_text']|e('html') }}</a>

koyan: Fixed and pushed.

classes/controller/blocks/class-p4bks-blocks-happypoint-controller.php
```diff
@@ -0,0 +1,119 @@
...
+		}
+
+		/**
+		 * Shortcode UI setup for the happypoint shortcode.
+		 * It is called when the Shortcake action hook `register_shortcode_ui` is called.
+		 */
+		public function prepare_fields() {
+			$fields = array(
+				array(
+					'label'       => esc_html__( 'Background', 'planet4-blocks' ),
```
dashaluna: Late escaping. I suggest doing `__()` instead of `esc_html__ ()` and let shortcode plugin to handle escaping on output.

And similar note for the following fields.

classes/controller/blocks/class-p4bks-blocks-happypoint-controller.php
```diff
@@ -0,0 +1,119 @@
...
+
+		/**
+		 * Callback for the shortcake_twocolumn shortcode.
+		 * It renders the shortcode based on supplied attributes.
+		 *
+		 * @param array $fields
+		 * @param string $content
+		 * @param string $shortcode_tag
+		 *
+		 * @return string
```
dashaluna: Docs: description of params and return values.

classes/controller/blocks/class-p4bks-blocks-happypoint-controller.php
```diff
@@ -0,0 +1,119 @@
...
+		public function prepare_template( $fields, $content, $shortcode_tag ) : string {
+
+			$fields = shortcode_atts( array(
+				'background'     	=> '',
+				'opacity' 		 	=> '30',
+				'boxout_title'    	=> '',
+				'boxout_descr'     	=> '',
+				'boxout_link_text' 	=> '',
+				'boxout_link_url'  	=> '',
+			), $fields, $shortcode_tag );
```
dashaluna: Alignments and redundant spaces, it should be aligned to `=>` of the longest array key.

classes/controller/blocks/class-p4bks-blocks-happypoint-controller.php
```diff
@@ -0,0 +1,119 @@
...
+
+			if (!is_numeric($fields['opacity'])) {
+				$fields['opacity'] = 70;
+			}
+			$opacity_reverse = number_format(($fields['opacity']/100), 1);
```
dashaluna: Minor - please space it out (the `if` statement and `$opacity_reverse`), re coding standards.

classes/controller/blocks/class-p4bks-blocks-happypoint-controller.php
```diff
@@ -0,0 +1,119 @@
...
+
+			$fields['background_html'] 		= wp_get_attachment_image( $fields['background'] );
+			$fields['background_src'] 		= wp_get_attachment_image_src( $fields['background'] );
+			$fields['opacity'] 				= $opacity_reverse;
+			$fields['boxout_title'] 		= wp_kses_post( $fields['boxout_title']);
+			$fields['boxout_descr'] 		= wp_kses_post( $fields['boxout_descr']);
+			$fields['boxout_link_text'] 	= wp_kses_post( $fields['boxout_link_text']);
+			$fields['boxout_link_url'] 		= esc_html( $fields['boxout_link_url']);
+
```
dashaluna: Alignments and redundant spaces, it should be aligned to `=` of the longest var.

classes/controller/blocks/class-p4bks-blocks-happypoint-controller.php
```diff
@@ -0,0 +1,119 @@
...
+
+			$data = ['fields' => $fields];
```
dashaluna: Spacing re Coding Standards

classes/controller/blocks/class-p4bks-blocks-happypoint-controller.php
```diff
@@ -0,0 +1,119 @@
...
+
+
+
```
dashaluna: Redundant new lines.

classes/controller/blocks/class-p4bks-blocks-happypoint-controller.php
dashaluna: I'm not sure about using escaping here. According to coding standards all the escaping should be done late on output. So I suggest, doing this in the twig template instead.

classes/controller/blocks/class-p4bks-blocks-happypoint-controller.php
```diff
@@ -0,0 +1,119 @@
...
+namespace P4BKS\Controllers\Blocks;
+
+if ( ! class_exists( 'P4BKS_Blocks_HappyPoint_Controller' ) ) {
+
+	class P4BKS_Blocks_HappyPoint_Controller extends P4BKS_Blocks_Controller {
+
+		/**
+		 *
+		 */
```
dashaluna: Missing docs.

classes/controller/blocks/class-p4bks-blocks-happypoint-controller.php
```diff
@@ -0,0 +1,119 @@
...
+				array(
+					'label'  => esc_html__( 'Boxout Link Url', 'planet4-blocks' ),
+					'attr'   => 'boxout_link_url',
+					'type'   => 'text',
+				),
+			);
+
+			/*
+			 * Define the Shortcode UI arguments.
+			 */
```
dashaluna: Use single line comment `// Define the Shortcode UI arguments.`

dashaluna:
Hello everyone, I will be doing code reviews from Human Made side. I'm a bit OCD with coding standards, so bear with me :)

classes/controller/blocks/class-p4bks-blocks-happypoint-controller.php
```diff
@@ -0,0 +1,119 @@
...
+					'attr'        => 'background',
+					'type'        => 'attachment',
+					'libraryType' => array( 'image' ),
+					'addButton'   => esc_html__( 'Select Background Image', 'planet4-blocks' ),
+					'frameTitle'  => esc_html__( 'Select Background Image', 'planet4-blocks' ),
+				),
+				array(
+					'label'  => esc_html__( 'Opacity % . Number between 1 and 100. If you leave it empty 30 will be used', 'planet4-blocks' ),
+					'attr'   => 'opacity',
+					'type'   => 'text',
```
dashaluna: It looks like shortcake allows `number` as type https://github.com/wp-shortcake/shortcake/wiki/Registering-Shortcode-UI I'd suggest using number here in this case.

koyan:
Comited new version with fixes for the suggested coding standards. Only missing is the description of the fields of the prepare_template function (inherited from shortcode_ui).

classes/controller/blocks/class-p4bks-blocks-happypoint-controller.php
koyan: Yes, implemented it in latest submit

classes/controller/blocks/class-p4bks-blocks-happypoint-controller.php
```diff
@@ -0,0 +1,114 @@
...
+				array(
+					'label' => __( 'Opacity % . Number between 1 and 100. If you leave it empty 30 will be used', 'planet4-blocks' ),
+					'attr'  => 'opacity',
+					'type'  => 'number',
+					'meta'  => array(
+						'data-test' => 30,
+					),
+				),
+				array(
+					'label' => esc_html__( 'Boxout Title', 'planet4-blocks' ),
```
dashaluna: Same comment as before re `esc_html__()` in this file for other fields declarations, should be used `__()` as we escape late.

Basically if code ins't around `echo()` or `s/printf()` - any output html being constructed that we just use translation functions.

classes/controller/blocks/class-p4bks-blocks-happypoint-controller.php
```diff
@@ -0,0 +1,114 @@
+<?php
+
+namespace P4BKS\Controllers\Blocks;
+
+if ( ! class_exists( 'P4BKS_Blocks_HappyPoint_Controller' ) ) {
+
+	class P4BKS_Blocks_HappyPoint_Controller extends P4BKS_Blocks_Controller {
+
+		/**
+		 * function to load the block and define its name
```
dashaluna: Coding Standards - Comments start with capital and end with a full stop.

classes/controller/blocks/class-p4bks-blocks-happypoint-controller.php
```diff
@@ -0,0 +1,114 @@
...
+		 * @param string $content The content of the post.
+		 * @param string $shortcode_tag The shortcode tag (shortcake_blockname).
+		 *
+		 * @return string The complete html of the block
+		 */
+		public function prepare_template( $fields, $content, $shortcode_tag ): string {
+
+			$fields = shortcode_atts( array(
+				'background'       => '',
+				'opacity'          => '30',
```
dashaluna: Why `'30'` is a string? Shouldn't it be a number?

classes/controller/blocks/class-p4bks-blocks-happypoint-controller.php
```diff
@@ -0,0 +1,114 @@
...
+				'boxout_title'     => '',
+				'boxout_descr'     => '',
+				'boxout_link_text' => '',
+				'boxout_link_url'  => '',
+			), $fields, $shortcode_tag );
+
+			if ( ! is_numeric( $fields['opacity'] ) ) {
+				$fields['opacity'] = 70;
```
dashaluna: In the code & comments above it says that default is 30, why it's 70 here?

classes/controller/blocks/class-p4bks-blocks-happypoint-controller.php
```diff
@@ -0,0 +1,114 @@
...
+					'label' => esc_html__( 'Boxout Link Url', 'planet4-blocks' ),
+					'attr'  => 'boxout_link_url',
+					'type'  => 'text',
+				),
+			);
+
+			// Define the Shortcode UI arguments.
+			$shortcode_ui_args = array(
+				'label'         => esc_html__( 'Happy Point', 'planet4-blocks' ),
+				'listItemImage' => '<img src="' . esc_url( plugins_url() . '/planet4-plugin-blocks/icons/happy_point.png' ) . '" />',
```
dashaluna: `esc_url()` is correct here, as HTML is being constructed.

classes/controller/blocks/class-p4bks-blocks-happypoint-controller.php
koyan: Fixed.

classes/controller/blocks/class-p4bks-blocks-happypoint-controller.php
koyan: Fixed.

classes/controller/blocks/class-p4bks-blocks-happypoint-controller.php
koyan: Fixed.

classes/controller/blocks/class-p4bks-blocks-happypoint-controller.php
koyan: Fixed.

classes/controller/blocks/class-p4bks-blocks-happypoint-controller.php
koyan: Fixed.

koyan:
@dashaluna All suggested changes commited. Please recheck.

PR #3, opened 2017-09-26
Feature/autoload static fix

27Bslash6:
I believe the .gitignore changes keep original intent and still enabled updating `vendor/composer/*` contents as well as `vendor/autoload.php`

Not sure about the extensive changes to `autoload_classmap.php`, and since they point to the squizlabs vendor directory, I'm requesting some advice as to how these updates should be managed.

Whatever happens, I believe we need to include `autoload_static.php` or future builds on PHP > 5.6 will continue to fail.


angtheod:
Hi Ray, you are right gitignore was not working as it was supposed to.
vendor should be vendor/*

As for the composer dump-autoload --optimize result,
did you run it without doing first composer update --no-dev? Because it looks like it included the class mappings for the dev packages (codesniffer).

27Bslash6:
Yes, you were correct I did not run --no-dev before the dump-autoload.  That's the reason for so much cruft.

Closing PR as .gitignore changes are implemented which resolved missing autoload_static.php

PR #6, opened 2017-10-02
Feature/1301 block content four column

kirdia:
Initial version of content four column block


kirdia:
Ok, I have reverted settings twig template to that of develop branch. Please ignore escaping in settings twig template as these will be fixed in another branch from @angtheod as @koyan suggested.

PR #7, opened 2017-10-03
Feature/1298 block subheader

koyan:
It misses the correct markup in the twig template (has not been delivered yet). 
But we could merge it with the temporary markup and create a new feature for the final markup


includes/blocks/subheader.twig
```diff
@@ -0,0 +1,16 @@
+{% block subheader %}
+	{% if ( fields ) %}
+		<div class="subheader-wrap">
+			<div class="container">
+				{% if ( fields.title ) %}
+					<div class="subheader-for-more">
+						<h2>{{ fields.title}}</h2>
```
angtheod: Just a space missing before the closing tags in lines 7,9
and a new line at the end of the file.
Other than that it looks good to me.

angtheod:
Missing vendor/composer/autoload_static.php change for adding the class mapping of the new class.
Composer > 1.1 uses this file instead of autoload_classmap.php

classes/controller/blocks/class-p4bks-blocks-subheader-controller.php
```diff
@@ -0,0 +1,76 @@
...
+					'label' => __( 'Description', 'planet4-blocks' ),
+					'attr'  => 'description',
+					'type'  => 'textarea',
+				),
+			);
+
+			// Define the Shortcode UI arguments.
+			$shortcode_ui_args = array(
+				'label'         => __( 'Subheader', 'planet4-blocks' ),
+				'listItemImage' => '<img src="' . esc_url( plugins_url() . '/planet4-plugin-blocks/icons/subheader.png' ) . '" />',
```
angtheod: We have already a directory for placing assets for the backend
admin/images is for the images. I should have noticed that on the happy_point branch as well but I missed it. Sorry for that. :/

PR #8, opened 2017-10-03
Feature/1299 block carouselsplit

kirdia:
Initial version of carousel split


koyan:
Closing pull request as it should be towards the develop, and not towards the master

PR #12, opened 2017-10-04
Initial version of media block

kirdia:


includes/blocks/media_block.twig
```diff
@@ -0,0 +1,13 @@
+{% block media %}
+    {% if ( fields ) %}
+
+        <div class="eager-wrap">
+            <div class="container">
+                {% if ( fields.attachment ) %}
+                    <img src="{{ fields.attachment }}"/>
+                {% endif %}
+            </div>
+
```
koyan: you have two empty lines (line 3 and 10) . A reason?

angtheod:
The autoload_static and autoload_classmap files? @kirdia

includes/blocks/media_block.twig
kirdia: No reason. Is not the final markup, it's a dummy one. I will remove them.

dashaluna:
Also, can you merge and fix conflicts :)

PR #17, opened 2017-10-04
Feature/1300 block articles

sagarsdeshmukh:
Initial version of articles block


classes/controller/blocks/class-p4bks-blocks-articles-controller.php
```diff
@@ -0,0 +1,99 @@
...
+			$args = array(
+				'numberposts' => $fields['article_count'],
+				'order'       => 'ASC',
+				'orderby'     => 'title',
+			);
+			$all_posts = wp_get_recent_posts( $args );
+
+			foreach ( $all_posts as $recent ) {
+				if ( has_post_thumbnail( $recent['ID'] ) ) {
+						$thumbnail_image     = get_the_post_thumbnail_url( $recent['ID'],'single-post-thumbnail' );
```
koyan: two tabs?

koyan: Also, why do you create a variable, just to assign it in the $recent['thumbnail']  ? Since you don't need the $thumbnail_image anywhere else in the function, just assign the value you want directly in the $recent['thumbnail']

classes/controller/blocks/class-p4bks-blocks-articles-controller.php
```diff
@@ -0,0 +1,99 @@
...
+						$recent['thumbnail'] = $thumbnail_image;
+				}
+				$recent['tags']     = wp_get_post_tags( $recent['ID'] );
+				$recent['category'] = get_the_category( $recent['ID'] );
+				$recent_posts[]     = $recent;
+			}
+
+			$data = [
+				'fields'        => array_map( 'wp_kses_post', $fields ),
```
koyan: Why do you escape all fields from $fields ? You are not showing them raw in the twig, so you don't need to do escaping at this stage. Just assign them. Without escaping

includes/blocks/articles.twig
```diff
@@ -0,0 +1,63 @@
+{% block articles %}
+	{% if ( recent_posts ) %}
+		<div class="news-section">
+				<div class="container">
```
koyan: Two tabs. You don't need two tabs

includes/blocks/articles.twig
```diff
@@ -0,0 +1,63 @@
...
+							<h3>{{ fields.article_heading }}</h3>
+						</div>
+						<div class="col-md-4 col-lg-3 ml-auto hidden-sm-down">
+							<a class="btn btn-medium btn-blue btn-block" href="#">READ ALL THE NEWS</a>
+						</div>
+						<div class="topicwise-news-section clearfix col-md-12">
+							<ul class="topicwise-news-list list-unstyled hidden-sm-down">
+								{% for key,value in recent_posts %}
+								<li class="media topicwise-news-list-itme {% if key == 1 %}my-5{% endif %}">
+									<img class="d-flex mr-3 topicwise-news-image" src="images/news-first.png" alt="topic wise news image placeholder">
```
koyan: Hardcoded image source and alt ? 
`src="images/news-first.png" alt="topic wise news image placeholder"`
Please use the correct image for the article you are showing

includes/blocks/articles.twig
```diff
@@ -0,0 +1,63 @@
...
+							<ul class="topicwise-news-list list-unstyled hidden-sm-up">
+								{% for value in recent_posts %}
+								<li class="topicwise-news-list-itme news-mobile-view {% if key == 1 %}my-5{% endif %}">
+									<div class="topicwise-news-body">
+									<div class="topicwise-news-tags">
+										<span class="news-topic-title">{% for category_data in value.category %} {{ category_data.name }} {% endfor %}</span>
+										<span class="news-topic-title-seprater"></span>
+										{% for tags_data in value.tags %}
+										<span class="news-topic-tag">#{{ tags_data.name }}</span>
+										{% endfor %}                <span class="news-topic-tag">#Water</span>
```
koyan: Probably you have left over from the demo the 
<span class="news-topic-tag">#Water</span>

Please clean it up.

classes/controller/blocks/class-p4bks-blocks-articles-controller.php
angtheod: I have to disagree here, escaping on the controller is always needed. You can not rely on some 3rd party plugin or a template engine for escape during presentation. @sagarsdeshmukh @koyan

includes/blocks/articles.twig
```diff
@@ -0,0 +1,63 @@
...
+									</div>
+									</div>
+										<img class="topicwise-news-image" src="images/news-first.png" alt="topic wise news image placeholder">
+								</li>
+								{% endfor %}
+							</ul>
+						</div>
+					</div>
+				</div>
+			</div>
```
koyan: Again, two tabs, so it is unclear which tags are closing which tags. Please fix the indentation of the whole file.

classes/controller/blocks/class-p4bks-blocks-articles-controller.php
```diff
@@ -0,0 +1,99 @@
...
+						'meta'  => array(
+							'placeholder' => __( 'Enter articles count', 'planet4-blocks' ),
+						),
+					),
+				);
+
+			// Define the Shortcode UI arguments.
+			$shortcode_ui_args = array(
+				'label'         => __( 'Articles', 'planet4-blocks' ),
+				'listItemImage' => 'dashicons-editor-table',
```
koyan: This is not wrong, but since the dashicons are used for other shortcodes , I suggest that you create an icon from the image of the design, and assign it like that: 
(line 94 of:)  https://github.com/greenpeace/planet4-plugin-blocks/blob/develop/classes/controller/blocks/class-p4bks-blocks-tasks-controller.php#L94

classes/controller/blocks/class-p4bks-blocks-articles-controller.php
```diff
@@ -0,0 +1,99 @@
...
+						'attr'  => 'article_heading',
+						'type'  => 'text',
+						'meta'  => array(
+							'placeholder' => __( 'Enter article heading', 'planet4-blocks' ),
+						),
+					),
+					array(
+						'label' => __( 'Article Count', 'planet4-blocks' ),
+						'attr'  => 'article_count',
+						'type'  => 'text',
```
koyan: Please change that to be a number and not text. Shortcode supports numbers.

classes/controller/blocks/class-p4bks-blocks-articles-controller.php
```diff
@@ -0,0 +1,99 @@
...
+		 *
+		 * @since 0.1.0
+		 *
+		 * @return string All the data used for the html.
+		 */
+		public function prepare_template( $fields, $content, $shortcode_tag ) : string {
+
+			// Get all posts with arguments.
+			$args = array(
+				'numberposts' => $fields['article_count'],
```
koyan: What happens if the editor has not provided a number for article_count ? 
Does it fail? does it show a default? What is the default? where do you define it? 
If it fails, please do a check before and assign a default number if none has been provided.

classes/controller/blocks/class-p4bks-blocks-articles-controller.php
koyan: @angtheod please check:  https://github.com/greenpeace/planet4-plugin-blocks/pull/1#discussion_r141008323
"According to coding standards all the escaping should be done late on output. So I suggest, doing this in the twig template instead."
If we are unsure if autoescape is enabled or not, we should be checking for it from the plugin core files, and if not, enable it from rhere.

dashaluna: I agree with @koyan here, escape late on output. Some of the main benefits:
1. You don't need to track code all the way up to see if it was previously escaped.
2. Escaping is done within context (different escaping needed for URL, HTML text, ect).

angtheod: @dashaluna I only said this because what if Timber or Twig introduce vulnerabilities that do not escape efficiently user input. But I know this is the proposed standard so yes it makes more sense.

P.S. Is number 2 really the case for us? I was wondering since we applied the autoescape strategy which by default applies the 'html' filter everywhere (not taking context into account) as far as I know. Perhaps, I am wrong about this though. And besides escaping with context can still happen earlier (WP context-aware functions esc_url, esc_attr, esc_html, ...)

classes/controller/blocks/class-p4bks-blocks-articles-controller.php
```diff
@@ -66,6 +66,7 @@ public function prepare_fields() {
 		 */
 		public function prepare_template( $fields, $content, $shortcode_tag ) : string {
 
+			$fields['article_count']   = ( isset( $fields['article_count'] ) && ! empty( $fields['article_count'] ) ) ? esc_attr( $fields['article_count'] ) : 3;
```
koyan: Why do you need to escape it? 
As we have said, we escape as late as possible, and there is no need to escape this at this stage.

classes/controller/blocks/class-p4bks-blocks-articles-controller.php
dashaluna: @angtheod well if we can't rely on Timber/Twig autoescape set up in the theme, then we need to explicitly escape each value in a template.

I don't think escaping earlier with WP native functions is a solution, as it's negates all valid benefits of escaping late.

As for #2 - that Twig doesn't provide context escaping - I think this a good indicator to step back and evaluate Twig usage altogether. From what I gather Twig's benefit is template inheritance but as far as I know it's not being used. Twig templates are still a mixture of HTML and code, that's the same as using the native approach of HTML + php, so we're not getting any pros from Twig on this front. Also, there are some cons of using Twig: need to learn new syntax, not as easily tested, doesn't use native reliable WP functionality (escaping is a really good example).

So if we're not making use of Twig's benefits and it introduces some obscurity and inflexibility, I would really question it's usage.

koyan: Oh, I want to invite @27Bslash6 into this, and have it resolved by a trial-by-combat, GoT style.

angtheod: > I don't think escaping earlier with WP native functions is a solution, as it's negates all valid benefits of escaping late.
Ye it wasn't clear enough what I wrote. Sorry for that. I wasn't really suggesting it as a solution, was more like questioning that applying context is a benefit of late-escaping since it can happen earlier as well.

My only worry about this is that practically  don't have any user input escape managing of our own, and that our strategy is to solely rely on timber and twig doing it for us. I am probably overreacting though.

koyan:
As discussed, please make the titles of the articles to be links (and the tags as well). Get the new markup from Ashish

PR #19, opened 2017-10-05

kirdia:
Renamed four column block to _static four column_ to avoid confusion with _content four column_ block


angtheod:
@kirdia 
Filename class-p4bks-blocks-static-four-column-controller.php
should be class-p4bks-blocks-staticfourcolumn-controller.php
to align with wp standards. Because it should be easy to convert from file name into class name and vice-versa with a custom autoloader.

PR #21, opened 2017-10-06
Feature/1323 Block: Covers.

angtheod:
It is ready for review but final markup is not yet delivered.


classes/controller/blocks/class-p4bks-blocks-covers-controllers.php
```diff
@@ -0,0 +1,127 @@
...
+
+				foreach ( $actions as $action ) {
+					$tags     = [];
+					$wp_tags = wp_get_post_tags( $action->ID );
+
+					if ( is_array( $wp_tags ) ) {
+						foreach ( $wp_tags as $wp_tag ) {
+							array_push( $tags, [
+								'name' => $wp_tag->name,
+								'href' => "/tag/$wp_tag->name",
```
koyan: How does relative urls work here? Most probably in our setup we will have www.domain.org/nro/ as the base url of each wp site. If your urls appear as relative in the page href="/tag/something"  then they will lead to www.domain.org/tag/something  , instead of www.domain.org/nro/tag/something  . This can be solved be using full urls. Please check how these appear in the front end and confirm.

angtheod: As it is now, the "/tag/something" is not relative (would be if it was "tag/something) but it still is based on the site's Url, so you are right there is a problem here if NRO Urls are like www.greenpeace.org/nro/.

includes/blocks/covers.twig
```diff
@@ -0,0 +1,42 @@
+{% block covers %}
+
+{% if ( fields and fields.covers ) %}
+    <div class="subheader">
```
koyan: Please change spaces to tabs for identation

includes/blocks/covers.twig
```diff
@@ -0,0 +1,42 @@
...
+    </div>
+    <div class="covers-block">
+        <div class="container">
+            <div class="row">
+                {% for cover in fields.covers %}
+                    <div class="col-lg-4 col-md-6">
+                        <div class="cover-card" style="background-image: url({{ cover.image }});">
+                        {% if ( cover.tags ) %}
+                            {% for tag in cover.tags %}
+                                <a class="cover-card-tag" href="{{ tag.href }}">#{{ tag.name }}</a>
```
koyan: Shouldn't we be using the tag slug instead of the name? The name can have spaces. I thought the slug was the one that should be used after the # symbol. ?

PR #23, opened 2017-10-10
Feature/1310 Block: Split Two Columns.

angtheod:


vendor/composer/autoload_static.php
```diff
@@ -39,10 +39,11 @@ class ComposerStaticInit356100170ff2882375cd04c4c76f9df3
         'P4BKS\\Controllers\\Blocks\\P4BKS_Blocks_CarouselSplit_Controller' => __DIR__ . '/../..' . '/classes/controller/blocks/class-p4bks-blocks-carouselsplit-controller.php',
         'P4BKS\\Controllers\\Blocks\\P4BKS_Blocks_Controller' => __DIR__ . '/../..' . '/classes/controller/blocks/class-p4bks-blocks-controller.php',
         'P4BKS\\Controllers\\Blocks\\P4BKS_Blocks_HappyPoint_Controller' => __DIR__ . '/../..' . '/classes/controller/blocks/class-p4bks-blocks-happypoint-controller.php',
+        'P4BKS\\Controllers\\Blocks\\P4BKS_Blocks_Split_Two_Columns_Controller' => __DIR__ . '/../..' . '/classes/controller/blocks/class-p4bks-blocks-split-two-columns-controller.php',
```
koyan: Shouldn't it be without underscores (everywhere)?

angtheod: Changed it to match the other cases.
The important thing is that the class name and the file name patterns match together.
So,
if class name: Split_Two_Columns then
file_name: splittwocolumns is not ok
file_name: split-two-columns is ok

cause we can then easily translate with code from one to the other, in case we create our own loader.

On a side note, the P4BKS_Blocks_ prefix in the class names is forgotten there since the early time that I had not introduced namespacing yet to the EN plugin and I forgot to remove it. It was then needed to avoid class name collisions, but now is redundant and hinders readability. I will make a PR to remove those and update the README file.
Changed it to match the other cases.
The important thing is that the class name and the file name patterns match together.
So,
if class name: Split_Two_Columns then
file_name: splittwocolumns is not ok
file_name: split-two-columns is ok

cause we can then easily translate with code from one to the other, in case we create our own loader.

@koyan @kirdia @sagarsdeshmukh 
On a side note, the P4BKS_Blocks_ prefix in the class names is forgotten there since the early time that I had not introduced namespacing yet to the EN plugin and I forgot to remove it. It was then needed to avoid class name collisions, but now is redundant and hinders readability. I will make a PR to remove those and update the README file.

classes/controller/blocks/class-p4bks-blocks-splittwocolumns-controller.php
```diff
@@ -0,0 +1,139 @@
...
+						'label' => get_the_title( $issue->ID ),
+					] );
+				}
+			}
+
+			$fields = [
+				[
+					'attr'        => 'select_issue',
+					'label'       => __( 'Select an Issue', 'planet4-blocks' ),
+					'description' => 'Associate this block to the Issue that it will talk about',
```
koyan: Localised message?

classes/controller/blocks/class-p4bks-blocks-splittwocolumns-controller.php
```diff
@@ -0,0 +1,139 @@
...
+
+				foreach ( $campaigns as $campaign ) {
+					$campaign_meta_data = get_post_meta( $issue_id );
+					$wp_tags = wp_get_post_tags( $campaign->ID );
+
+					if ( is_array( $wp_tags ) ) {
+						array_push( $campaigns_splits, [
+							'tag'         => [
+								'name' => $wp_tags[0]->name,
+								'href' => "$site_url/tag/" . $wp_tags[0]->name,
```
koyan: get_tag_link

classes/controller/blocks/class-p4bks-blocks-splittwocolumns-controller.php
```diff
@@ -0,0 +1,137 @@
...
+
+			$campaigns = get_posts( [
+				'post_type'     => 'page',
+				'category_name' => 'campaigns',
+			] );
+
+			$campaigns_splits = [];
+			if ( $campaigns ) {
+				foreach ( $campaigns as $campaign ) {
+					$campaign_meta_data = get_post_meta( $issue_id );
```
kirdia: This is probably a typo, because you have already pulled get_post_meta( $issue_id ), maybet you intended to use campaign id

classes/controller/blocks/class-p4bks-blocks-splittwocolumns-controller.php
```diff
@@ -0,0 +1,137 @@
...
+
+		/**
+		 * Shortcode UI setup for the shortcode.
+		 * It is called when the Shortcake action hook `register_shortcode_ui` is called.
+		 */
+		public function prepare_fields() {
+
+			$issues = get_posts( [
+				'post_type'     => 'page',
+				'category_name' => 'issues',
```
johnbillion: This only pulls in the latest 10 posts by default. You might want to add in the `posts_per_page` argument to set it to a higher value, or set it to `-1` to see all issues.

classes/controller/blocks/class-p4bks-blocks-splittwocolumns-controller.php
```diff
@@ -0,0 +1,137 @@
...
+
+					if ( is_array( $wp_tags ) && $wp_tags ) {
+						$campaigns_splits[] = [
+							'tag'         => [
+								'name' => $wp_tags[0]->name,
+								'href' => get_tag_link( $wp_tags[0]->term_id ),
+							],
+							'description' => $campaign_meta_data['p4_description'][0],
+							'image'       => get_the_post_thumbnail_url( $campaign->ID ),
+							'link_text'   => __( 'SUPPORT THIS CAMPAIGN', 'planet4-blocks' ),
```
johnbillion: For accessibility reasons, you may want to use capitalised lower case text here and make it appear uppercase using CSS.

classes/controller/blocks/class-p4bks-blocks-splittwocolumns-controller.php
```diff
@@ -0,0 +1,137 @@
...
+							'link_url'    => get_post_permalink( $campaign->ID ),
+						];
+					}
+				}
+			}
+
+			$data = [
+				'issue' => [
+					'title'       => null === $issue_meta_data['p4_title'][0] ? get_the_title( $issue_id ) : $issue_meta_data['p4_title'][0],
```
johnbillion: A nice feature was introduced in PHP 7 called the null coalesce operator. You can now use this shorthand for null checks in ternary conditionals:

```php
'title' => $issue_meta_data['p4_title'][0] ?? get_the_title( $issue_id ),
```

angtheod:
@johnbillion changes commited, please re-review

PR #25, opened 2017-10-13
Remove class prefixes

angtheod:
Remove the P4BKS_ prefixes in class names that were used before introducing namespacing and were forgotten. These prefixes were redundant, were hindering readability and making class names too long.

Example:
**Class Name**
'P4BKS\Controllers\Blocks\CarouselSplit_Controller' => 'P4BKS\Controllers\Blocks\CarouselSplit_Controller',

**File Name**
class-p4bks-blocks-carouselsplit-controller.php =>
class-carouselsplit-controller.php

*Note, if merged, it will create conflicts to the already existing PRs.


koyan:
The timing you have done that is unfortunate: Since not all blocks have been merged into develop yet, you have removed this only from some blocks, and not from all. This will result that when we merge the other blocks in to develop , having two naming conventions (and also a very big number of conflicts). 

You will have to redo this job after all blocks have been merged to develop.

PR #29, opened 2017-10-18
Feature/1269 block tasks

kirdia:
Latest changes to block tasks


classes/controller/blocks/class-p4bks-blocks-tasks-controller.php
```diff
@@ -30,56 +30,80 @@ public function load() {
...
+			];
+
 			// This block will have 4 different columns with same fields.
-			$fields = [];
 			for ( $i = 1; $i < 5; $i++ ) {
 				$field = [
 					[
-						'label' => __( 'Title', 'planet4-blocks' ),
+						// translators: placeholder needs to represent the ordinal of the task/column, eg. 1st, 2nd etc.
+						'label' => sprintf( __( 'Task #%s Title', 'planet4-blocks' ), $i ),
```
koyan: Like in the content four column, please try to minimise the number of things (Especially similar) that we ask the translators to translate.
'label' => __( 'Task') . ' ' . $i . ': ' .  __( 'Title'),
Could solve this (resulting in 9 + 1 = 10 translation strings, instead of 36 that your code results).

angtheod:
Generally, when we simply want to add elements to an indexed array, it is less memory and cpu intensive
to use array_push() instead of array_merge() ( [benchmark )](https://gist.github.com/kalimatas/6e8d3d455443922820074396329cfadb)
or even better as [php suggests](http://php.net/array_push) simply assign values
`$fields[] = $field`

@kirdia @koyan @sagarsdeshmukh

kirdia:
> Generally, when we simply want to add elements to an indexed array, it is less memory and cpu intensive
to use array_push() instead of array_merge() ( benchmark )
or even better as php suggests simply assign values to avoid the overhead of a function call (I should do that as well)
$fields[] = $field

Yeah, that's a good point. Merge is not really needed for this, it just makes it cleaner :)

PR #32, opened 2017-10-18
Feature/1323 Block: Covers

angtheod:
Added optional Tag selection, so that Editors can display only Action…


classes/controller/blocks/class-p4bks-blocks-covers-controllers.php
```diff
@@ -39,6 +40,13 @@ public function prepare_fields() {
 						'data-plugin' => 'planet4-blocks',
 					],
 				],
+				[
+					'attr'        => 'select_tag',
+					'label'       => __( 'Select a Tag', 'planet4-blocks' ),
+					'description' => 'Associate this block to Actions that have a specific Tag',
```
koyan: Why not localised message?

koyan:
Looks good 👍

PR #34, opened 2017-10-19
Feature/1300 block articles

koyan:


classes/controller/blocks/class-p4bks-blocks-articles-controller.php
```diff
@@ -75,18 +75,47 @@ public function prepare_template( $fields, $content, $shortcode_tag ) : string {
...
+						$tags_data['href'] = "$site_url/tag/$wp_tag->slug";
+						$tags[]            = $tags_data;
+					}
+				}
+
+				$recent['tags'] = $tags;
+				$wp_categories  = get_the_category( $recent['ID'] );
+
+				$categories = array();
+				if ( is_array( $wp_categories ) ) {
```
angtheod: get_the_category() returns always an array().
If no categories found then it will return an empty array, so we need to check if it is empty, not if it is an array
`if ( $wp_categories ) {`

includes/blocks/articles.twig
```diff
@@ -19,17 +19,17 @@
 								<div class="media-body topicwise-article-body">
 								<div class="topicwise-article-tags">
 									{% if ( value.category ) %}
-										<a href="#" class="article-topic-title">{% for category_data in value.category %} {{ category_data.name }} {% endfor %}</a>
+										{% for category_data in value.category %} <a href="{{ category_data.href }}" class="article-topic-title"> {{ category_data.name }}</a>{% endfor %}
```
angtheod: make this more readable by placing html on different line

classes/controller/blocks/class-p4bks-blocks-articles-controller.php
```diff
@@ -75,18 +75,47 @@ public function prepare_template( $fields, $content, $shortcode_tag ) : string {
...
 				}
-				$recent['tags']     = wp_get_post_tags( $recent['ID'] );
-				$recent['category'] = get_the_category( $recent['ID'] );
-				$recent_posts[]     = $recent;
+
+				$wp_tags = wp_get_post_tags( $recent['ID'] );
+
+				$tags = array();
+
+				if ( is_array( $wp_tags ) ) {
```
angtheod: Same here (sorry). I need to change this as well in my blocks.
Just to clarify, this wouldn't cause any errors if left like this, it's just makes sure it does not enter the if statement at all if array is empty, instead of entering and then not doing any loop. It is very trivial change, but still...

angtheod: `if ( is_array( $wp_tags ) && $wp_tags ) {`
Just add check if its not empty array ( I need to add this as well )

PR #36, opened 2017-10-19
Give the whole Object instead of an ID to WP's get_ functions to avoi…

angtheod:
…d unnecessary work performed by them.


koyan:
Looks good 👍

PR #37, opened 2017-10-19
Give the whole Object instead of an ID to WP's get_ functions to avoi…

koyan:
…d unnecessary work performed by them.


angtheod:
1. The $site_url line is not needed any more
2. Do same for functions (has_post_thumbnail, get_the_post_thumbnail_url, get_post_thumbnail_id)
3. In line 76: wp_get_recent_posts can return false which would cause an error on the following loop. So, needs an if ( $all_posts ) before the loop. (Sorry, I missed that on the last PR)

angtheod:
Updated previous comment

koyan:
Removing the ID from the get_the_post_thumbnail_url  and  get_post_thumbnail_id resulted in no image url and the same id always. I am leaving those two as they are.

PR #39, opened 2017-10-23
Feature/1315 block carousel header

kirdia:


classes/controller/blocks/class-p4bks-blocks-carousel-header-controller.php
```diff
@@ -0,0 +1,158 @@
...
+		 * @param array  $attributes    Defined attributes array for this shortcode.
+		 * @param string $content       Content.
+		 * @param string $shortcode_tag Shortcode tag name.
+		 *
+		 * @return string Returns the compiled template.
+		 */
+		public function prepare_template( $attributes, $content, $shortcode_tag ) : string {
+
+			$attributes_temp = [];
+			for ( $i = 1; $i < 5; $i++ ) {
```
sagarsdeshmukh: Ideally, loop should not be hardcoded here. What if author added data for only first 3?

classes/controller/blocks/class-p4bks-blocks-carousel-header-controller.php
```diff
@@ -0,0 +1,158 @@
...
+					"description_$i"  => wpautop( $attributes[ "description_$i" ] ),
+					"image_$i"        => $attributes[ "image_$i" ],
+					"link_text_$i"    => $attributes[ "link_text_$i" ],
+					"link_url_$i"     => $attributes[ "link_url_$i" ],
+				];
+				$attributes_temp = array_merge( $attributes_temp, $temp_array );
+			}
+			$attributes = shortcode_atts( $attributes_temp, $attributes, $shortcode_tag );
+
+			for ( $i = 1; $i < 5; $i++ ) {
```
sagarsdeshmukh: Same like above - loop should not be hardcoded

includes/blocks/carousel_header.twig
```diff
@@ -0,0 +1,42 @@
+{% block carousel_header %}
+
+    {% if ( fields ) %}
+
+        <div class="four-coloum">
+            <div class="container">
+                <div class="row">
+                    {% for i in 1..4 %}
```
sagarsdeshmukh: There can be unwanted loops if there is no data for `i=3` or `i=4` etc. Ideally, here also loops should not be hardcoded

PR #40, opened 2017-10-24
Feature/1318 block media video

sagarsdeshmukh:


classes/controller/blocks/class-p4bks-blocks-mediavideo-controller.php
```diff
@@ -0,0 +1,81 @@
+<?php
+
+namespace P4BKS\Controllers\Blocks;
+
+if ( ! class_exists( 'P4BKS_Blocks_Mediavideo_Controller' ) ) {
+
+	class P4BKS_Blocks_Mediavideo_Controller extends P4BKS_Blocks_Controller {
```
dashaluna: Might be worth having `Mediavideo` as `Media_Video` throughout the file, it seems to be easier to read.

dashaluna: Oh, I've just noticed there is other controller with class name including `HappyPoint`. So I guess to stay consistent we can rename `P4BKS_Blocks_Mediavideo_Controller` to `P4BKS_Blocks_MediaVideo_Controller`.
But I'd still suggest using `media_video` inside the PHP code.

classes/controller/blocks/class-p4bks-blocks-mediavideo-controller.php
```diff
@@ -0,0 +1,81 @@
...
+
+
+		/**
+		 * Override this method in order to give your block its own name.
+		 */
+		public function load() {
+			$this->block_name = 'mediavideo';
```
dashaluna: Use `media_video` for better readability?

dashaluna:
I don't wanna merge - in case there is order preference for that. I've added new label `ready for merge` and will assign that when a PR is approved. So you can go ahead and merge when convenient.

Let's see how this workflow works :)

dashaluna:
@sagarsdeshmukh there is a note from `codacy` - worth looking into fixing that one too 👍

koyan:
Ok, no problem I will continue doing the merges

PR #48, opened 2017-10-26
Feature/1310 Block: Split Two Columns - update

angtheod:
Fixes issue with selecting also pages other than the Issue pages (e.g. Take Action pages which have as parent page the Issue page).


koyan:
Looks good 👍

PR #51, opened 2017-10-27
Changed content four column based on new requirements

kirdia:


classes/controller/blocks/class-p4bks-blocks-contentfourcolumn-controller.php
```diff
@@ -150,33 +62,41 @@ public function prepare_fields() {
...
+			if ( $query->have_posts() ) {
+
+				$posts = $query->get_posts();
+
+				foreach ( $posts as $post ) {
+
+					$post->alt_text  = '';
+					$post->thumbnail = '';
+
+					if ( has_post_thumbnail( $post ) ) {
```
sagarsdeshmukh: Better if we used post ID here. So if `( has_post_thumbnail( POST_ID_HERE ) ) {` @koyan - Do you think same issue like article thumbnails can occur if we use this ?

koyan: Have you tested it and it doesn't work as it should? 
From the documentation : https://developer.wordpress.org/reference/functions/has_post_thumbnail/  I see that this function takes either the ID or the post object itself.

sagarsdeshmukh: Yes both will work. My question is which is more better :)

koyan: If both work, then using the post is better. If you look at the code of the function, if you pass it the ID it calls the post object itself. If you pass it the post object, then you save the application from one more big object loading function.

PR #52, opened 2017-10-28
Get covers only for Take Action pages that have page ACT as parent. R…

angtheod:
…eplace wp_get_recent_posts() with get_posts(). Use direct assignment for indexed array instead of array_push(). Remove unecessary get_site_url() call.


koyan:
This will do for now. For after the prototype we will have to do the following change: https://jira.greenpeace.org/browse/PLANET-1411  (NOT FOR NOW)

PR #56, opened 2017-10-30
1413 Post edit screen preview for shortcake blocks

tcrsavage:
- Adds a conditional to P4BKS_Blocks_Controller shortcode registration so we can render differently on preview
- Adds an iframe render method to render the shortcode via iframe in markup
- Adds a iframe ajax callback to render the markup defined by the preview iframe with html wrapper markup + site frontend scripts and styles

This is pretty hacky but looks to me like the best way to drop in the admin preview capability without either generating new admin script and style sheets specifically for these previews or having to modify existing functionality in the each registered block's prepare_template method.

see https://jira.greenpeace.org/browse/PLANET-1413

**Before**

screen shot 2017-10-30 at 16 07 59

**After**
screen shot 2017-10-30 at 16 07 15

Will require some continued work/testing during development of these modules and frontend styling to give good compatibility

### DIFF
--- a/classes/controller/blocks/class-p4bks-blocks-controller.php
+++ b/classes/controller/blocks/class-p4bks-blocks-controller.php
@@ -135,20 +135,20 @@ if ( ! class_exists( 'P4BKS_Blocks_Controller' ) ) {
 
 			$request = add_query_arg(
 				array_merge( $fields,
-					[
+					array(
 						'_tag'     => $shortcode_tag,
 						'_content' => $content,
 						'_post_id' => get_the_ID(),
-						'_nonce'    => wp_create_nonce( 'p4bks_preview_render_' . $this->block_name ),
+						'_nonce'   => wp_create_nonce( 'p4bks_preview_render_' . $this->block_name ),
 						'action'   => 'p4bks_preview_render_' . $this->block_name
-					]
+					)
 				),
 				admin_url( 'admin-ajax.php' )
 			);
 
 			ob_start();
 			?>
-				<iframe width="100%" src="<?php echo esc_url( $request ) ?>" onload="this.style.height = this.contentWindow.document.body.scrollHeight + 'px';"></iframe>
+			<iframe width="100%" src="<?php echo esc_url( $request ) ?>" onload="this.style.height = this.contentWindow.document.body.scrollHeight + 'px';"></iframe>
 			<?php
 			return ob_get_clean();
 		}
@@ -160,6 +160,11 @@ if ( ! class_exists( 'P4BKS_Blocks_Controller' ) ) {
 		 * via $this->prepare_template along with wrapper html and enqueued frontend styles and scripts
 		 */
 		public function prepare_admin_preview() {
+
+			// Shortcode UI not callable
+			if ( ! is_callable( 'Shortcode_UI', 'get_shortcode' ) ) {
+				exit;
+			}
 
 			$tag                 = sanitize_text_field( $_GET['_tag'] );
 			$content             = wp_kses_post( $_GET['content'] );
@@ -190,8 +195,8 @@ if ( ! class_exists( 'P4BKS_Blocks_Controller' ) ) {
 				setup_postdata( $post );
 				// @codingStandardsIgnoreEnd
 			}
-			?>
 
+			?>
 			<html>
 				<head>
 					<?php do_action( 'wp_head' ); ?>
--- a/classes/controller/blocks/class-p4bks-blocks-controller.php
+++ b/classes/controller/blocks/class-p4bks-blocks-controller.php
@@ -166,32 +166,33 @@ if ( ! class_exists( 'P4BKS_Blocks_Controller' ) ) {
 				exit;
 			}
 
-			$tag                 = sanitize_text_field( $_GET['_tag'] );
-			$content             = wp_kses_post( $_GET['content'] );
-			$post_id             = isset( $_GET['_post_id'] ) && is_numeric( $_GET['_post_id'] ) ? absint( $_GET['_post_id'] ) : false;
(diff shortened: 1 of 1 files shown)


tcrsavage:
@angtheod can I hand this to you for review?

classes/controller/blocks/class-p4bks-blocks-controller.php
```diff
@@ -101,6 +122,99 @@ public function shortcode_ui_register_shortcodes() {
...
+			}
+
+			// Setup postdata incase it's needed during shortcode render
+			if ( get_post( $post_id ) ) {
+				// @codingStandardsIgnoreStart
+				global $post;
+				$post = get_post( $post_id );
+				setup_postdata( $post );
+				// @codingStandardsIgnoreEnd
+			}
```
angtheod: Hello @tcrsavage ,
could we call here $post = get_post( $post_id );
before the if block and then check if ( $post ) to avoid double get_post() call?

classes/controller/blocks/class-p4bks-blocks-controller.php
```diff
@@ -101,6 +122,99 @@ public function shortcode_ui_register_shortcodes() {
...
+			$content             = wp_kses_post( $_GET['content'] );
+			$post_id             = isset( $_GET['_post_id'] ) && is_numeric( $_GET['_post_id'] ) ? absint( $_GET['_post_id'] ) : false;
+			$shortcode_object    = \Shortcode_UI::get_instance()->get_shortcode( $tag );
+			$shortcode_attrs     = is_array( $shortcode_object ) && is_array( $shortcode_object['attrs'] ) ? $shortcode_object['attrs'] : [];
+			$shortcode_attr_keys = wp_list_pluck( $shortcode_attrs, 'attr' );
+
+			// Don't do anything if an invalid nonce
+			if ( ! wp_verify_nonce( $_GET['_nonce'], 'p4bks_preview_render_' . $this->block_name ) ) {
+				exit;
+			}
```
angtheod: @tcrsavage 
could we call tthe wp_verify_nonce function here earlier (like after the is_callable() check) to avoid unnecessary code in case this check fails?

classes/controller/blocks/class-p4bks-blocks-controller.php
```diff
@@ -101,6 +122,99 @@ public function shortcode_ui_register_shortcodes() {
...
+		public function prepare_admin_preview() {
+
+			// Shortcode UI not callable
+			if ( ! is_callable( 'Shortcode_UI', 'get_shortcode' ) ) {
+				exit;
+			}
+
+			$tag                 = sanitize_text_field( $_GET['_tag'] );
+			$content             = wp_kses_post( $_GET['content'] );
+			$post_id             = isset( $_GET['_post_id'] ) && is_numeric( $_GET['_post_id'] ) ? absint( $_GET['_post_id'] ) : false;
```
angtheod: @tcrsavage 
this part is completely ok, but just wanted to ask you if you could
`filter_input( INPUT_GET, '_post_id', FILTER_VALIDATE_INT );` so that WordPress.VIP Coding Standards stops complaining that we access super global directly? It will also validate that it isset and that it is_numeric.

classes/controller/blocks/class-p4bks-blocks-controller.php
```diff
@@ -101,6 +122,99 @@ public function shortcode_ui_register_shortcodes() {
...
 
 		/**
+		 * Output markup of an iframe to render shortcode when previewing in admin edit screen.
+		 *
+		 * We need to load through iframe to enqueue frontend styles without breaking admin ui
+		 *
+		 * @param $fields
+		 * @param $content
+		 * @param $shortcode_tag
+		 * @return string
```
angtheod: * @param array  $fields Array of fields that are to be used in the template.
* @param string $content The content of the post.
* @param string $shortcode_tag The shortcode tag (shortcake_two_columns).

I know these were missing from the prepare_template() function as well, so if you could add those to it as well would be great  :)

classes/controller/blocks/class-p4bks-blocks-controller.php
```diff
@@ -101,6 +122,99 @@ public function shortcode_ui_register_shortcodes() {
...
+
+			$request = add_query_arg(
+				array_merge( $fields,
+					array(
+						'_tag'     => $shortcode_tag,
+						'_content' => $content,
+						'_post_id' => get_the_ID(),
+						'_nonce'   => wp_create_nonce( 'p4bks_preview_render_' . $this->block_name ),
+						'action'   => 'p4bks_preview_render_' . $this->block_name
+					)
```
angtheod: @tcrsavage 
please lets use the short syntax for arrays, I know we need to apply this everywhere... I am only using the array() syntax for hooking method callbacks within a class because it looks better there...

Also, could you please add a comma on the last item of the array as well?

classes/controller/blocks/class-p4bks-blocks-controller.php
```diff
@@ -101,6 +122,99 @@ public function shortcode_ui_register_shortcodes() {
...
+				),
+				admin_url( 'admin-ajax.php' )
+			);
+
+			ob_start();
+			?>
+			<iframe width="100%" src="<?php echo esc_url( $request ) ?>" onload="this.style.height = this.contentWindow.document.body.scrollHeight + 'px';"></iframe>
```
angtheod: @tcrsavage 
Inline php statements should end with a semicolon

angtheod:
@tcrsavage 
:+1: Works beautifully, very nice addition. I am sure Editors will much appreciate this.
Just a few minor changes I suggested.

tcrsavage:
Thanks @angtheod !

Missed a couple of items here before @koyan trigger merged :p Will reopen to merge those in too

PR #58, opened 2017-10-31
Feature/1320 block campaign thumbnail

sagarsdeshmukh:


classes/controller/blocks/class-p4bks-blocks-campaignthumbnail-controller.php
```diff
@@ -0,0 +1,105 @@
...
+		 * @param string $content This is the post content.
+		 * @param string $shortcode_tag The shortcode block of campaign thumbnail.
+		 *
+		 * @since 0.1.0
+		 *
+		 * @return string All the data used for the html.
+		 */
+		public function prepare_template( $fields, $content, $shortcode_tag ) : string {
+
+			$category_data  = get_the_category();
```
tcrsavage: A relatively minor one but get_the_category is returning term objects which have been remapped into very old data structure. `get_the_terms` or `wp_get_post_categories` would be better for this. `wp_get_post_categories` defaults to returning IDs only which would simplify these conditionals a bit

classes/controller/blocks/class-p4bks-blocks-campaignthumbnail-controller.php
```diff
@@ -0,0 +1,105 @@
...
+			$category_id    = ( isset( $category_data[0]->cat_ID ) && ! empty( $category_data[0]->cat_ID ) ) ? $category_data[0]->cat_ID : '';
+
+			if( isset( $category_data[0]->cat_ID ) && ! empty( $category_data[0]->cat_ID ) ) {
+				$args         = array( 'numberposts' => '1', 'category' => $category_id );
+			} else {
+				$args         = array( 'numberposts' => '1' );
+			}
```
tcrsavage: - Mixing of `[]` and `array()` syntax. We should start enforcing `[]`. 
- Pulling out variable initialisation from the conditional makes the code a bit more flexible

```
$args  = [ 'numberposts' => '1' ]

if (  is_array( $category_data ) &&  ! empty( $category_data[0]->cat_ID ) ) {
    $args['cagegory'] = $category_id 
}

```

classes/controller/blocks/class-p4bks-blocks-campaignthumbnail-controller.php
```diff
@@ -0,0 +1,105 @@
...
+		 * @since 0.1.0
+		 *
+		 * @return string All the data used for the html.
+		 */
+		public function prepare_template( $fields, $content, $shortcode_tag ) : string {
+
+			$category_data  = get_the_category();
+			$category_id    = ( isset( $category_data[0]->cat_ID ) && ! empty( $category_data[0]->cat_ID ) ) ? $category_data[0]->cat_ID : '';
+
+			if( isset( $category_data[0]->cat_ID ) && ! empty( $category_data[0]->cat_ID ) ) {
```
tcrsavage: `isset` check is redundant. Should probably check that you got an array back instead. With `get_the_category()` you're guaranteed an array back. wp_get_post_categories can return a WP_Error object in some cases.

classes/controller/blocks/class-p4bks-blocks-campaignthumbnail-controller.php
```diff
@@ -0,0 +1,105 @@
...
+				$args         = array( 'numberposts' => '1', 'category' => $category_id );
+			} else {
+				$args         = array( 'numberposts' => '1' );
+			}
+
+			$recent_posts   = wp_get_recent_posts( $args );
+			$recent_post_id = ( isset( $recent_posts[0]['ID'] ) && ! empty( $recent_posts[0]['ID'] ) ) ? $recent_posts[0]['ID'] : '';
```
tcrsavage: Redundant isset check

classes/controller/blocks/class-p4bks-blocks-campaignthumbnail-controller.php
```diff
@@ -0,0 +1,105 @@
...
+			if( ! empty( $recent_post_id ) ) {
+				$post_tags    = get_the_tags( $recent_post_id );
+				$post_tags    = array_slice($post_tags, 0, 3);
+			}
+
+			foreach( $post_tags as $tag ) {
+				$tags           = array();
+				$tags['name']   = $tag->name;
+				$tags['slug']   = $tag->slug;
+				$tags['href']   = get_tag_link( $tag );
```
tcrsavage: I'd give `tags` a more appropriate variable name and initialise in 1 go here. 

```
$tag_remapped  = [ 
    'name' => $tag->name;
    'slug'   => $tag->slug;
    'href'   => get_tag_link( $tag );
]
```

classes/controller/blocks/class-p4bks-blocks-campaignthumbnail-controller.php
sagarsdeshmukh: Hi @tcrsavage : I will need to get the categories of the current active page here. With `wp_get_post_categories` and `get_the_terms` - do you think is this possible ? because I tried with `wp_get_post_categories` and it gives me blank array and  for `get_the_terms` - I need to pass and ID i guess. Can you please confirm.

tcrsavage: HI @sagarsdeshmukh `wp_get_post_categories` Yeah you'll need to pass the ID in there, `$cat_ids = wp_get_post_categories( get_the_id() );`

PR #62, opened 2017-11-06
Disabled autostart. Disabled relative video suggestions

koyan:


kirdia:
Looks ok!

27Bslash6:
Wouldn't it make sense for these parameters to be options configurable in the backend?

koyan:
We didn't have such a requirement. The requirement was to disable both autostart and relative videos. If an editor still wants, they can embed a video the traditional way

PR #63, opened 2017-11-06
Initial version of take action boxout block

kirdia:


classes/controller/blocks/class-p4bks-blocks-takeactionboxout-controllers.php
```diff
@@ -0,0 +1,124 @@
...
+		/**
+		 * Shortcode UI setup for the shortcode.
+		 * It is called when the Shortcake action hook `register_shortcode_ui` is called.
+		 */
+		public function prepare_fields() {
+
+			// Get the id of the ACT page. We need this to get the children posts/pages of the ACT Page.
+			$arguments = [
+				'post_type'     => 'page',
+				'post_name__in' => [ 'act', 'ACT', 'Act' ],
```
koyan: I am a little reluctant with that. For the prototype we are ok, but NROs are going to have different names for the same page. 
Maybe remove this restriction?

koyan:
Looks good 👍

PR #65, opened 2017-11-07
Updated markup for content four column. Added load more functionality…

kirdia:
… for content four column additional columns


includes/blocks/content_four_column.twig
```diff
@@ -1,46 +1,63 @@
...
+			{% if ( posts|length > 4 ) %}
+
+				<div class="row mt-5 load-more-button-div">
+					<div class="col-md-4 col-lg-3 col-xl-3 mb-5">
+						<button class="btn btn-block btn-medium btn-small btn-primary"
+						        onclick="$( '.c4-row-hidden' ).first().css( 'display', '' ).removeClass('c4-row-hidden');
+						        if ($( '.c4-row-hidden' ).size() == 0) {
+						        	$('.load-more-button-div').css('display', 'none');
+						        }"
+						>
```
angtheod: @kirdia Hmm I am not a fan of inline js to be honest.
Do you agree to take this outside of the markup?
It makes markup more readable and its easier to spot a js error

includes/blocks/content_four_column.twig
```diff
@@ -1,46 +1,63 @@
...
+				{% endif %}
+
+			{% endfor %}
+
+			{% if ( posts|length > 4 ) %}
+
+				<div class="row mt-5 load-more-button-div">
+					<div class="col-md-4 col-lg-3 col-xl-3 mb-5">
+						<button class="btn btn-block btn-medium btn-small btn-primary"
+						        onclick="$( '.c4-row-hidden' ).first().css( 'display', '' ).removeClass('c4-row-hidden');
```
27Bslash6: I was going to agree with Angelos here, for these reasons (not sure why the comment was deleted?) ... This was the comment text:

+1 @angtheod 
Plus, if there's a network delay loading jQuery and the user clicks the button there would be a javascript error that may impact all other global-scope JS interactions.
Prefer to avoid any onclick, and either go with native .addEventListener or jQuery $.on() - ideally delegated events on parent elements, to also interact with elements added after domready.
See: https://api.jquery.com/on/#direct-and-delegated-events

angtheod: Hey @27Bslash6 
I re-added my comment as a Review comment

27Bslash6: Ahah, too many cooks :)

angtheod: In case of jQuery's .on() its usually good practice to first remove  any previously attached handlers with .off(), like:
`$(selector).off('event').on('event', function() {});`

kirdia: Yeah I agree. I added there temporarily. I will move it to the js files

PR #66, opened 2017-11-07
Deprecate Controller's protected  property and replace it with a cons…

angtheod:
…tant along with the use of Late Static Binding. This way it is simpler to retrieve the name of a block (e.g. Covers::BLOCK_NAME ) and at the same time there is no longer requirement to override the load() method when developing a new block.


angtheod:
It should have been this way from the start. Remembered this when finding a case inside the Tag page template where I needed the names of some blocks and had to create new object's and then call load() to get dynamically the block of a name... which makes no sense...

This also removes the requirement to override method load().

koyan:
Looks good 👍

PR #69, opened 2017-11-08
Added functionality to select multiple tags for content four column. …

kirdia:
…Removed link from post excerpt in column body


koyan:
Approving it and merging it, with the following follow up task: 
https://jira.greenpeace.org/browse/PLANET-1498

PR #72, opened 2017-11-08
Fixed query using multiple tags for content four column. Do not show …

kirdia:
…any html output if no post is associated with the selected tags.


koyan:
Looks good 👍

PR #73, opened 2017-11-08
Feature/1482 take action boxout

kirdia:
Added background image for take action boxout. Truncate page excerpt. Add link to take action page


koyan:
Looks good 👍

PR #76, opened 2017-11-10
Fix Select an Issue dropdown to show only Issue options. Categories t…

angtheod:
…hat are issues need to be childnen of category <Issues>. In the right side, show Campaing name instead of slug. Also, change issue link on left side to pretty link.

### DIFF
--- a/classes/controller/blocks/class-p4bks-blocks-splittwocolumns-controller.php
+++ b/classes/controller/blocks/class-p4bks-blocks-splittwocolumns-controller.php
@@ -20,16 +20,16 @@ if ( ! class_exists( 'P4BKS_Blocks_SplitTwoColumns_Controller' ) ) {
 		 */
 		public function prepare_fields() {
 
-			$issues = get_posts( [
-				'post_type'     => 'page',
-				'category_name' => 'issues',
-				'post_parent'   => 0,
-				'numberposts'   => -1,
+			$categories = get_categories( [
+				'parent'  => get_cat_ID( 'Issues' ),            // Issue categories needs to be children of category Issues.
+				'orderby' => 'name',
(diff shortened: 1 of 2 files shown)


classes/controller/blocks/class-p4bks-blocks-splittwocolumns-controller.php
```diff
@@ -20,16 +20,16 @@ class P4BKS_Blocks_SplitTwoColumns_Controller extends P4BKS_Blocks_Controller {
 		 */
 		public function prepare_fields() {
 
-			$issues = get_posts( [
-				'post_type'     => 'page',
-				'category_name' => 'issues',
-				'post_parent'   => 0,
-				'numberposts'   => -1,
+			$categories = get_categories( [
+				'parent'  => get_cat_ID( 'Issues' ),            // Issue categories needs to be children of category Issues.
```
koyan: Keeping a note (also in RT : https://jira.greenpeace.org/browse/PLANET-1411 ) This will need to change to a theme setting before we can go to different languages.

koyan:
Looks good 👍

classes/controller/blocks/class-p4bks-blocks-splittwocolumns-controller.php
angtheod: Yes, or if we provide another solution for relating Category and Issue page (like add a field inside the Categories to connect it to the Issue page) that could solve both of these problems. @koyan

koyan: ok, please suggest your solution in the jira issue. When we get to that point, we should have it available for consideration

PR #78, opened 2017-11-10
Fix Two columns block disparity. Removed default values from button t…

angtheod:
…ext. Buttons now disappear if text or link are not set. Added xss protection for raw output fields.


koyan:
Looks good 👍

PR #86, opened 2017-11-21
Feature/1323 block covers

angtheod:
Make Covers block show 6 (or 4 on mobile) covers initially. Make dynamic the number of covers per row based on HTTP_USER_AGENT so that the Load more button works nicer on mobile devices. This should be done in frontend but solves issue for now.


angtheod:
If Twig [extension](https://github.com/suncat2000/MobileDetectBundle) is worth looking

koyan:
Looks good 👍

PR #88, opened 2017-11-22

kirdia:
Changed indentation from spaces to tabs, that's why all lines are changes.
The only real change is adding empty alt attribute to img tag.


koyan:
Looks good 👍

PR #94, opened 2017-11-29
removing unneeded fields from happy point block

koyan:


classes/controller/blocks/class-p4bks-blocks-happypoint-controller.php
```diff
@@ -36,32 +36,12 @@ public function prepare_fields() {
...
-					'label' => __( 'Boxout Link Text', 'planet4-blocks' ),
-					'attr'  => 'boxout_link_text',
-					'type'  => 'text',
-				),
-				array(
-					'label' => __( 'Boxout Link Url', 'planet4-blocks' ),
-					'attr'  => 'boxout_link_url',
-					'type'  => 'text',
-				),
 				array(
```
kirdia: I think we should use one style of array declaration.
Since in most cases we use `$array = []`, we should keep it consistent.
This style is introduced in php 5.3. We have as base php 7.0 so it won't be a problem.

classes/controller/blocks/class-p4bks-blocks-happypoint-controller.php
```diff
@@ -89,11 +69,7 @@ public function prepare_template( $fields, $content, $shortcode_tag ): string {
 			$fields = shortcode_atts( array(
 				'background'          => '',
 				'opacity'             => 30,
-				'boxout_title'        => '',
-				'boxout_descr'        => '',
-				'boxout_link_text'    => '',
-				'boxout_link_url'     => '',
-				'mailing_list_iframe' => ''
+				'mailing_list_iframe' => '',
```
kirdia: Just a small notice for more clarity. I think it is better to create the array outside the function call, to be more clear.
`$fields = shortcode_atts( $temp_fields, $fields, $shortcode_tag );

koyan: Again: This PR is for fixing one bug issue, not for changing the way this works. Open an issue and we will fix it with a dedicated branch/PR

classes/controller/blocks/class-p4bks-blocks-happypoint-controller.php
koyan: I agree, but this PR is for fixing a very specific issue, not for changing the array usage in this code base. Lets keep issues seperate.

classes/controller/blocks/class-p4bks-blocks-happypoint-controller.php
kirdia: Ok, besides these, it looks good 👍

PR #95, opened 2017-11-29
Remove unnecessary class name prefixes that were there before we add …

angtheod:
…namespacing. Few minor coding standard fixes.


koyan:
Looks good 👍

PR #100, opened 2017-12-05

kirdia:
Added escaping for static four column fields in twig template.


koyan:
Looks good 👍

PR #104, opened 2017-12-05
Added latest markup for content four column.

kirdia:


includes/blocks/content_four_column.twig
```diff
@@ -7,51 +7,46 @@
...
-								{% endif %}
-							</div>
-
+						<div class="col-md-4 col-lg-3 col-xl-3">
+							<a href="{{ post.permalink }}">
+								<div class="four-coloum-content-wrap clearfix">
+									<div class="four-coloum-content-info">
+										<div class="four-coloum-content-symbol">
+											{% if ( post.thumbnail ) %}
+												<img src="{{ post.thumbnail|raw }}" alt="{{ post.alt_text }}">
```
koyan: Why do we pass this raw? The url is not having any strange characters in it

kirdia: Yeah, good catch. It is not needed, I will remove it.

koyan:
Looks good 👍

PR #107, opened 2017-12-07
Add isset checks to the Carousel Header block to avoid fatal PHP error.

fklein-lu:
Empty fields inside of the Shortcake UI are not added to the shortcode attributes. This means that these fields are then not set inside of the `$attributes` array in the `prepare_template()` method. This leads to a fatal PHP error.


koyan:
Please do not commit to master. Commit to develop. We do not work from master

fklein-lu:
Right. If there was any kind of documentation on the project workflows, this whole dance could have been avoided.

PR #109, opened 2017-12-07
Make the Header Carousel use the predined large image size.

fklein-lu:


includes/blocks/carousel_header.twig
```diff
@@ -22,7 +22,7 @@
 							{% if ( attribute( fields, 'header_'~i ) and attribute( fields, 'image_'~i ) ) %}
 								<div class="{{ loop.first ? 'carousel-item active': 'carousel-item' }}">
 									{% if ( attribute( fields, 'image_'~i ) ) %}
-										<img src="{{ attribute( fields, 'image_'~i ) }}" alt="{{ attribute( fields, 'image_'~i~'_alt' )}}">
+										<img src="{{ attribute( fields, 'image_'~i )|resize(1920,775)  }}" alt="{{ attribute( fields, 'image_'~i~'_alt' )}}">
```
koyan: I see that you use twig's resizing function. Why resize two times? Wasn't the whole point to decide which of the two resizing functionalities we would use?

koyan:
Closing as we decided to move without jetpack

PR #111, opened 2017-12-07
Use direct call to `jetpack_photon_url()` to generate resized image.

fklein-lu:
As an alternative to #109, this PR uses a direct call to `jetpack_photon_url()` to retrieve the resized image.

If this works, then the URL of the image should start with `https://i1.wp.com/` or similar, which is the domain of the WordPress.com image CDN.


koyan:
Closing as we decided to move without jetpack

PR #113, opened 2017-12-08
Feature/1315 select focus point of background image

koyan:
New approach on both resizing, and passing the image's correct focus point.
First: The editor can select (if they don't like the default) a focus point for an image they upload:
screen shot 2017-12-08 at 16 10 30

Second: The image tag is generated with the correct sizes (as we will define them), and also a new parameter is added for the background-position. 
screen shot 2017-12-08 at 16 09 09

As a third step (not done by me, as I don't want to touch the js of the presentation layer), the front end js developer needs to pick up this new parameter and pass it on to the second image. 
to-be-done

(I have tested it by manually changing the value of that line).


classes/controller/blocks/class-carouselheader-controller.php
```diff
@@ -39,6 +39,23 @@ public function prepare_fields() {
 						'addButton'   => __( 'Select Image', 'shortcode-ui' ),
 						'frameTitle'  => __( 'Select Image', 'shortcode-ui' ),
 					],
+					[
+						// translators: placeholder needs to represent the ordinal of the image, eg. 1st, 2nd etc.
+						'label'       => '<img src="' . esc_url( plugins_url() . '/planet4-plugin-blocks/admin/images/grid_9.png' ) . '" />' . sprintf( __( 'Select focus point for %s image', 'planet4-blocks' ),  $i ),
```
fklein-lu: `plugins_url()` accepts a `$path` argument.

fklein-lu: `$i` needs to be escaped.

classes/controller/blocks/class-carouselheader-controller.php
```diff
@@ -39,6 +39,23 @@ public function prepare_fields() {
 						'addButton'   => __( 'Select Image', 'shortcode-ui' ),
```
fklein-lu: The `shortcode-ui` text domain should be changed.

koyan: Correct. Did not think that the related plugin may not be translated in all the languages we are using. Since we have used it (wrong and extensively) throughout the plugin (and not only in this piece of code), and in order not to mix contexts, I am opening a different issue for it (so that it gets fixed everywhere).  https://github.com/greenpeace/planet4-plugin-blocks/issues/116

classes/controller/blocks/class-carouselheader-controller.php
koyan: $i is generated by a for statement a few lines above. It is not user input, it is not coming from the database. Why do we need to escape it?

PR #114, opened 2017-12-11
Carousel blocks changes - pre and next arrow should not display when …

sagarsdeshmukh:
…there is single image

### DIFF
--- a/includes/blocks/carousel_split.twig
+++ b/includes/blocks/carousel_split.twig
@@ -38,11 +38,11 @@
 					{% if fields|length > 1 %}
 						<a class="carousel-control-prev" href="#carousel-wrapper" role="button" data-slide="prev">
 							<span class="carousel-control-prev-icon" aria-hidden="true"><i></i></span>
-							<span class="sr-only">{{ __( 'Previous', domain ) }}</span>
+							<span class="sr-only">{{ __( 'Previous' ) }}</span>
 						</a>
 						<a class="carousel-control-next" href="#carousel-wrapper" role="button" data-slide="next">
 							<span class="carousel-control-next-icon" aria-hidden="true"><i></i></span>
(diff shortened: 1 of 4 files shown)


includes/blocks/carousel_split.twig
```diff
@@ -33,14 +35,16 @@
...
-						<span class="sr-only">Previous</span>
-					</a>
-					<a class="carousel-control-next" href="#carousel-wrapper" role="button" data-slide="next">
-						<span class="carousel-control-next-icon" aria-hidden="true"><i></i></span>
-						<span class="sr-only">Next</span>
-					</a>
+					{% if fields|length > 1 %}
+						<a class="carousel-control-prev" href="#carousel-wrapper" role="button" data-slide="prev">
+							<span class="carousel-control-prev-icon" aria-hidden="true"><i></i></span>
+							<span class="sr-only">{{ __( 'Previous', domain ) }}</span>
```
koyan: I think you need to remove the "domain" from there. 
@kirdia please confirm (Since you did the same investigation a couple days ago).

includes/blocks/carousel_split.twig
```diff
@@ -33,14 +35,16 @@
...
+						</a>
+						<a class="carousel-control-next" href="#carousel-wrapper" role="button" data-slide="next">
+							<span class="carousel-control-next-icon" aria-hidden="true"><i></i></span>
+							<span class="sr-only">{{ __( 'Next', domain ) }}</span>
```
koyan: Same here with the "domain"

includes/blocks/carousel_split.twig
kirdia: Yes, right. no need for that. Since you are making changes, please remove also text domain. Is not need in this case.

kirdia:
Looks good

PR #115, opened 2017-12-11
PLANET-1552 Update the campaign thumbnail markup

sagarsdeshmukh:


includes/blocks/campaign_thumbnail.twig
```diff
@@ -1,20 +1,22 @@
...
-								<div class="thumbnail-large">
-								{% if ( tag.image[0] ) %}
-									<img src="{{ tag.image[0] }}" alt="{{ value.alt_text }}" >
-								{% endif %}
-								<a href="{{ tag.href }}">#{{ tag.name }}</a>
-								</div>
+								<a href="{{ tag.href }}">
+									<div class="thumbnail-large">
+										{% if ( tag.image[0] ) %}
+											<img src="{{ tag.image[0] }}" alt="{{ value.alt_text }}" >
```
angtheod: @sagarsdeshmukh Should we add a default value for the alt text or is it always set?

includes/blocks/campaign_thumbnail.twig
```diff
@@ -1,20 +1,22 @@
...
 	{% if ( fields.tags ) %}
-		<div class="container">
-			<div class="campaign-thumbnail-block">
+		<div class="campaign-thumbnail-block">
+			<div class="container">
 				{% if ( fields.title ) %}
 					<h2>{{ fields.title }}</h2>
 				{% endif %}
 				<div class="row thumbnail-largeview-container">
 					{% if ( fields.tags ) %}
```
kirdia: This check is unneeded since it is already checked in the 2nd line. I know it wasn't introduced now, but by removing it will decrease the nesting levels.

includes/blocks/campaign_thumbnail.twig
sagarsdeshmukh: @angtheod  Default it is blank

PR #119, opened 2017-12-13
Add img tag with srcset in twig templates.

kirdia:
We are converting img tags in twig templates with img tags with srcet attribute, that will enable us serving different images based on the device.

### DIFF
--- a/includes/blocks/campaign_thumbnail.twig
+++ b/includes/blocks/campaign_thumbnail.twig
@@ -1,14 +1,14 @@
 {% block campaign_thumbnail %}
 	{% if ( fields.tags ) %}
-		<div class="container">
-			<div class="campaign-thumbnail-block">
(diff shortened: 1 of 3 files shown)


includes/blocks/campaign_thumbnail.twig
```diff
@@ -1,23 +1,28 @@
 {% block campaign_thumbnail %}
 	{% if ( fields.tags ) %}
-		<div class="campaign-thumbnail-block">
```
sagarsdeshmukh: @kirdia I see this one is the old markup which I updates yesterday. Can you please confirm again
https://github.com/greenpeace/planet4-plugin-blocks/pull/115

kirdia: Yes, you are right. I made the changes with the latest markup. Thanx for pointing out!

PR #121, opened 2017-12-15
Feature/fix retina image generation

27Bslash6:
Uses wp_calculate_image_srcset() to generate retina sizes. This hooks into low-lying Wordpress functionality, respected by other image manipulation libraries - notably WP-Stateless.

This also fixes:
- the retina image generation which was non-functional before (it only created a series of same-sized images in the srcset)
- Displays 'large' image by default instead of 'full' image. 

By merging this we'll re-introduce CDN image delivery on docker.p4.greenpeace.org, where with only the raw Timber resize code the images were being delivered locally.

This commit is a first-pass, representative of how the code in all other blocks should be generating image sizes.  

Help wanted:
- modifying other plugin block classes and templates
- registering standard image sizes with [add_image_size](https://developer.wordpress.org/reference/functions/add_image_size/)
- removing hardcoded '1024,768' maximum image size in line `wp_calculate_image_srcset(['1024', '768'], ... ` https://github.com/greenpeace/planet4-plugin-blocks/pull/121/commits/a723a3ae3d817fc68bdf504c92c0af7b152b5a38#diff-7ff25cf52abdb904511dd5ae2e7bd224R157

### DIFF
--- a/classes/controller/blocks/class-articles-controller.php
+++ b/classes/controller/blocks/class-articles-controller.php
@@ -92,6 +92,7 @@ if ( ! class_exists( 'Articles_Controller' ) ) {
 				foreach ( $all_posts as $recent ) {
 					$recent['alt_text']  = '';
 					$recent['thumbnail'] = '';
+					$recent['author']    = get_the_author_meta( 'display_name', $recent['post_author'] );
 
 					if ( has_post_thumbnail( $recent['ID'] ) ) {
 						$recent['thumbnail'] = get_the_post_thumbnail_url( $recent['ID'], 'single-post-thumbnail' );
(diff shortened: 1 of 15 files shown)


koyan:
> Help wanted:

>    modifying other plugin block classes and templates
>    registering standard image sizes with add_image_size
>    removing hardcoded '1024,768' maximum image size in line 
We will take these over. Any reason we should not merge this to start with?

koyan:
After discussion with Ray, we are going to do small incremental changes . We are going to merge this PR, and hung all the related PRs under the issue https://github.com/greenpeace/planet4-plugin-blocks/issues/122

PR #123, opened 2017-12-15
Replaced p4-page-type select box with checkboxes.

kirdia:
Replaced p4-page-type select box with checkboxes in content four column block.
Made Publication post type selected by default.
Fixed bug when both p4-page-type and tag were selected.

### DIFF
--- a/classes/controller/blocks/class-contentfourcolumn-controller.php
+++ b/classes/controller/blocks/class-contentfourcolumn-controller.php
@@ -25,6 +25,28 @@ if ( ! class_exists( 'ContentFourColumn_Controller' ) ) {
 		 */
 		public function prepare_fields() {
 
+			$checkboxes                 = [];
+			$planet4_article_type_terms = get_terms(
+				[
+					'hide_empty' => false,
+					'orderby'    => 'name',
+					'taxonomy'   => 'p4-page-type',
+				]
+			);
+
+			// Construct a checkbox for each p4-page-type.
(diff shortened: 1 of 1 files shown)


classes/controller/blocks/class-contentfourcolumn-controller.php
```diff
@@ -25,6 +25,28 @@ class ContentFourColumn_Controller extends Controller {
...
+				]
+			);
+
+			// Construct a checkbox for each p4-page-type.
+			if ( ! empty( $planet4_article_type_terms ) ) {
+				foreach ( $planet4_article_type_terms as $term ) {
+					$checkboxes [] = [
+						'attr'        => 'p4_page_type_' . str_replace( '-', '_', $term->slug ),
+						'label'       => $term->name . ' Posts',
+						'description' => 'Use Posts that belong to ' . $term->name . ' type to populate the content of this block',
```
koyan: Shouldn't this be a localised message?

kirdia: But p4-page-type terms would be in English, right? If the custom taxonomy have different names in each installation then it does have a meaning to be translated.

koyan: Indeed, it seems like we have not translated them to start with (in the definition of the page). Ok, lets leave it as it is for now, and if we have a request we translate all of them (both here and in the master-theme where we define them).

PR #147, opened 2018-01-02
Change css class names for load more button functionality.

kirdia:
Changed content four column and covers block to work with changed load more button functionality.

### DIFF
--- a/includes/blocks/content_four_column.twig
+++ b/includes/blocks/content_four_column.twig
@@ -46,7 +46,7 @@
 			{% if ( posts|length > 4 ) %}
 				<div class="row mt-5 load-more-button-div">
 					<div class="col-md-4 col-lg-3 col-xl-3 mb-5">
-						<button class="btn btn-block btn-medium btn-secondary btn-load-more-blocks">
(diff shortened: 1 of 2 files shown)


includes/blocks/content_four_column.twig
```diff
@@ -46,7 +46,7 @@
 			{% if ( posts|length > 4 ) %}
 				<div class="row mt-5 load-more-button-div">
 					<div class="col-md-4 col-lg-3 col-xl-3 mb-5">
-						<button class="btn btn-block btn-medium btn-secondary btn-load-more">
+						<button class="btn btn-block btn-medium btn-secondary btn-load-more-blocks">
```
angtheod: @kirdia 
Do you think it'd be better to name the class functionality-wise (e.g. btn-load-more-click)
instead of context-wise, in order to make sense to use it in other places apart from inside blocks.

Respectively, we could name the class for the load more on scroll functionality btn-load-more-scroll or even btn-load-more-click-scroll if it is supposed to have both. What you think?

kirdia: Yes, it is a good idea. I will make the needed changes to master-theme also.

PR #166, opened 2018-01-09
Remove next css class assignment in second slide for styling reasons.

kirdia:
Remove assignment of 'next' class to 2nd slide during initial loading as it results in 2nd slide's header & subheader text blending with 1st slide's header & subheader (opacity: 1 set by css).
Sliding and wrapping functionality is not impaired.


angtheod:
Spreadsheet issue 152

PR #168, opened 2018-01-09
Add background gradient to cover cards

comzeradd:
This is a design fix to improve readability of tags and titles on action cards


comzeradd:
Yes, good catch. I amend the commit.

PR #169, opened 2018-01-11
happy point label change for opacity

sagarsdeshmukh:
PLANET-1585 label changed for block

### DIFF
--- a/classes/controller/blocks/class-happypoint-controller.php
+++ b/classes/controller/blocks/class-happypoint-controller.php
@@ -29,7 +29,7 @@ if ( ! class_exists( 'HappyPoint_Controller' ) ) {
 					'frameTitle'  => __( 'Select Background Image', 'planet4-blocks' ),
 				],
 				[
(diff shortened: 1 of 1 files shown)


classes/controller/blocks/class-happypoint-controller.php
```diff
@@ -29,7 +29,7 @@ public function prepare_fields() {
 					'frameTitle'  => __( 'Select Background Image', 'planet4-blocks' ),
 				],
 				[
-					'label' => __( 'Opacity % . Number between 1 and 100. If you leave it empty 30 will be used', 'planet4-blocks' ),
+					'label' => __( '<i>Use an overlay to fade the image back. Use a number between 1 and 100,<br /> the higher the number, the more faded the image will look. If you leave this <br/> empty, the default of 30 will be used.</i>', 'planet4-blocks' ),
```
koyan: Please change it to : "We use an overlay". 
(Add the word "We" at the beggining), to make more sense.

PR #170, opened 2018-01-11
Bug/added srcset to split two column images

koyan:


includes/blocks/split_two_columns.twig
```diff
@@ -31,9 +31,13 @@
 		<div class="split-two-column-item item--right">
 			{% if ( campaign.image ) %}
 				{% if ( campaign.focus ) %}
-					<div class="split-two-column-item-image" style="background-image:url({{ campaign.image }}); background-position:{{ campaign.focus }};"></div>
+					<div class="split-two-column-item-image" style="background-position:{{ campaign.focus }};">
+						<img src="{{ campaign.image }}" srcset="{{ campaign.srcset }}">
```
sagarsdeshmukh: Do you think we need alt attribute here ?

koyan: Correct. Adding it now

includes/blocks/split_two_columns.twig
```diff
@@ -31,9 +31,13 @@
 		<div class="split-two-column-item item--right">
 			{% if ( campaign.image ) %}
 				{% if ( campaign.focus ) %}
```
angtheod: Should we have a default value for campaign.focus? If image is set but focus is not?

koyan: No, if none is given by the editor, the center is used by default by the css

angtheod: oh yea.. and there is an else to that if... didn't see it.. sorry

PR #171, opened 2018-01-12
Feature/split carousel design changes

koyan:
taken from : greenpeace/planet4-presentation-layer#176


kirdia:
Looks good!
Are the changes to articles.twig intended?

koyan:
Yes, they are intended. I wrote in the relevant git commit.

PR #187, opened 2018-01-22
Added optional fields in split two column block

kirdia:
[Issue 1650](https://jira.greenpeace.org/browse/PLANET-1650)
Added optional fields for issue title, issue description, issue link, issue image and campaign image.
Changed select_tag to a simple select field because of an [issue](https://github.com/wp-shortcake/shortcake/issues/775) with select2 fields in shortcake (v. 0.7.3)

### DIFF
--- a/classes/controller/blocks/class-splittwocolumns-controller.php
+++ b/classes/controller/blocks/class-splittwocolumns-controller.php
@@ -120,12 +120,12 @@ if ( ! class_exists( 'SplitTwoColumns_Controller' ) ) {
 					'description' => __( '(Optional)', 'planet4-blocks' ),
 				],
 				[
-					'label'       => __( 'Issue Image', 'shortcode-ui' ),
+					'label'       => __( 'Issue Image', 'planet4-blocks' ),
 					'attr'        => 'issue_image',
 					'type'        => 'attachment',
 					'libraryType' => [ 'image' ],
-					'addButton'   => __( 'Select Image for Issue', 'shortcode-ui' ),
-					'frameTitle'  => __( 'Select Image for Issue', 'shortcode-ui' ),
+					'addButton'   => __( 'Select Image for Issue', 'planet4-blocks' ),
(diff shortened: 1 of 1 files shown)


classes/controller/blocks/class-splittwocolumns-controller.php
```diff
@@ -102,14 +163,30 @@ public function prepare_fields() {
...
+					'label' => __( 'Campaign button link', 'planet4-blocks' ),
 					'attr'  => 'button_link',
 					'type'  => 'url',
 					'meta'  => [
 						'placeholder' => __( 'Enter button link', 'planet4-blocks' ),
 					],
 					'description' => __( '(Optional)', 'planet4-blocks' ),
 				],
+				[
+					'label'       => __( 'Campaign Image', 'shortcode-ui' ),
```
sagarsdeshmukh: We need to use same text domain throughout the plugin blocks. We must use `planet4-blocks` instead of `shortcode-ui` throughout

classes/controller/blocks/class-splittwocolumns-controller.php
```diff
@@ -56,44 +56,105 @@ public function prepare_fields() {
...
+					'label' => __( 'Issue link path', 'planet4-blocks' ),
+					'attr'  => 'issue_link_path',
+					'type'  => 'url',
+					'meta'  => [
+						'placeholder' => __( 'Enter link path', 'planet4-blocks' ),
+					],
+					'description' => __( '(Optional)', 'planet4-blocks' ),
+				],
+				[
+					'label'       => __( 'Issue Image', 'shortcode-ui' ),
```
angtheod: Shouldn't the text domain be 'planet4-blocks' ?
Other than that looks ok. I am not sure what is the new desired functionality but code looks ok.

classes/controller/blocks/class-splittwocolumns-controller.php
kirdia: Good catch. I updated the code.

PR #197, opened 2018-01-25
PLANET-1660  current post should not display in post article block

sagarsdeshmukh:


classes/controller/blocks/class-articles-controller.php
```diff
@@ -85,6 +85,7 @@ public function prepare_template( $fields, $content, $shortcode_tag ) : string {
 			$fields['article_count']  = ( ! empty( $fields['article_count'] ) ) ? $fields['article_count'] : 3;
 			$fields['read_more_text'] = $fields['read_more_text'] ?? __( 'READ ALL THE NEWS', 'planet4-blocks' );
 			$fields['read_more_link'] = ( ! empty( $fields['read_more_link'] ) ) ? $fields['read_more_link'] : '/?s=&orderby=post_date&f[ctype][Post]=3';
+			$exclude_post_id          = $fields['post_id'] ?? '';
```
angtheod: maybe its better to type cast this to int since post__not_in expects an array of integers.
`(int) $fields['post_id'] ?? '';`

PR #208, opened 2018-02-06
PLANET-1763 On other than tag page, In article block-read more link s…

sagarsdeshmukh:
…hould lead to search page preselected with current page categories

### DIFF
--- a/classes/controller/blocks/class-articles-controller.php
+++ b/classes/controller/blocks/class-articles-controller.php
@@ -100,9 +100,9 @@ if ( ! class_exists( 'Articles_Controller' ) ) {
 			// On other than tag page, read more link should lead to search page-preselected with current page categories.
 			if ( '' == $tag_id ) {
 				$category_filter = '';
-				if ( $category_id_array ) {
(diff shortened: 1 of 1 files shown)


classes/controller/blocks/class-articles-controller.php
```diff
@@ -97,6 +97,18 @@ public function prepare_template( $fields, $content, $shortcode_tag ) : string {
 				$category_id_array[] = $category->term_id;
 			}
 
+			// On other than tag page, read more link should lead to search page-preselected with current page categories.
+			if ( '' == $tag_id ) {
+				$category_filter = '';
+				if ( $category_id_array ) {
+					foreach ( $category_id_array as $category_id ) {
+						$category_filter .= '&f[cat][' . get_category($category_id)->name . ']=' . $category_id;
```
kirdia: You can use `$post_categories ` variable for the loop, since it already contains WP_Term objects and it will prevent making another query to db  ( `get_category()` )

sagarsdeshmukh: Yes I just fix it, please review.

PR #217, opened 2018-02-16
Revert "switched to large from specific size"

koyan:
Reverts greenpeace/planet4-plugin-blocks#216


koyan:
Resubmitting, after having fixed the color issue (using imageMagick)

PR #219, opened 2018-02-16
PLANET-1768 Post - display articles related to current post tags

sagarsdeshmukh:


classes/controller/blocks/class-articles-controller.php
```diff
@@ -146,6 +148,17 @@ public function prepare_template( $fields, $content, $shortcode_tag ) : string {
...
 			}
 
+			// For post, display related article based on current post tags.
+			if ( 'post' === get_post_type() ) {
+				if ( $post_tags ) {
+					$tag_id_array = [];
+					foreach ( $post_tags as $tag ) {
+						$tag_id_array[] = $tag->term_id;
+					}
+					$args['tag_id'] = implode( ',', $tag_id_array );
```
kirdia: I have not checked it but will 'tag_id' work for multiple tag ids?
I was thinking if 'tag_id' is for a single tag id and the 'tag__in' should be used for this purpose.

sagarsdeshmukh: It is fixed. Please check.

PR #224, opened 2018-02-21
Feature/1586 add focus point

ashish173:
Using image tag for happy point now.


comzeradd:
Code looks good. But since this has some backend changes too I'd feel more comfortable if @kirdia also takes a look on it before we merge.

classes/controller/blocks/class-happypoint-controller.php
```diff
@@ -66,6 +83,7 @@ public function prepare_template( $fields, $content, $shortcode_tag ): string {
 			$shortcode_atts_pairs =[
 				'background'          => '',
 				'opacity'             => 30,
+				"focus_image" => $attributes[ "focus_image" ] ?? '',
```
kirdia: Please assign a default focus_image value (probably 'center center') if it is not defined to avoid breaking css in the twig template.

PR #241, opened 2018-03-16
PLANET-1942 instead of category select a subpage of explore

koyan:
This negates the need for PLANET-1889

The need to assign pages to a category was because we needed a way to select a category for the split-two-column, and show the associated page.

But in reality we don't need to select a category. We need to select a page.

So, by changing the logic of the block, and selecting only pages which are children of the "Explore" page, we achieve the same result, without needing any more the requirement that "A page and a category must have the same name"

### DIFF
--- a/classes/controller/blocks/class-splittwocolumns-controller.php
+++ b/classes/controller/blocks/class-splittwocolumns-controller.php
@@ -32,27 +32,26 @@ if ( ! class_exists( 'SplitTwoColumns_Controller' ) ) {
 				[ 'value' => 'right bottom',  'label' => __( '9 - Bottom Right', 'planet4-blocks' ) ],
 			];
 
-			$option_values     = get_option( 'planet4_options' );
-			$issue_category_id = $option_values['issues_parent_category'] ?? '';
-			$categories        = [];
-			if ( 0 !== absint( $issue_category_id ) ) {
-				$categories   = get_categories( [
-					'parent'    => $issue_category_id,                  // Get the dynamic id of the 'Issue' category.
-					'orderby'   => 'name',
-					'order'     => 'ASC',
-				] );
-			}
+			$option_values   = get_option( 'planet4_options' );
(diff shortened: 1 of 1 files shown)


classes/controller/blocks/class-splittwocolumns-controller.php
```diff
@@ -32,27 +32,26 @@ public function prepare_fields() {
...
-				] );
-			}
+			$option_values   = get_option( 'planet4_options' );
+			$options         = [];
+			$explore_page_id = $option_values['explore_page'] ?? '';
+
+			$args = array(
+				'sort_order'   => 'asc',
+				'sort_column'  => 'post_title',
+				'hierarchical' => 1,
```
angtheod: we could skip this one as it is 1 by default

classes/controller/blocks/class-splittwocolumns-controller.php
```diff
@@ -32,27 +32,26 @@ public function prepare_fields() {
...
+				'child_of'     => $explore_page_id,
+				'post_type'    => 'page',
+				'post_status'  => 'publish'
+			);
+			$pages = get_pages($args);
```
angtheod: spaces missing after and before parenthesis

classes/controller/blocks/class-splittwocolumns-controller.php
```diff
@@ -32,27 +32,26 @@ public function prepare_fields() {
...
+			$options         = [];
+			$explore_page_id = $option_values['explore_page'] ?? '';
+
+			$args = array(
+				'sort_order'   => 'asc',
+				'sort_column'  => 'post_title',
+				'hierarchical' => 1,
+				'child_of'     => $explore_page_id,
+				'post_type'    => 'page',
+				'post_status'  => 'publish'
```
angtheod: comma missing after last key/value pair

PR #246, opened 2018-03-26
PLANET-1868 - Adjust four column block to make it responsive

comzeradd:
The change here is that I wrapped the `four-column-wrap` column under an `if` statement to make sure we actually have content to display. 

As an example see the [About Us](https://dev.p4.greenpeace.org/international/about/) page, where we have content for 3 columns but we render 4, preventing the block from expanding to full width.

### DIFF
--- a/includes/blocks/static_four_column.twig
+++ b/includes/blocks/static_four_column.twig
@@ -7,39 +7,41 @@
 				{% endif %}
 				<div class="row">
 					{% for i in 1..4 %}
-						<div class="col-md-6 col-lg four-column-wrap">
-							{% set link_url = fields['link_url_'~i] %}
-							{% if ( fields['attachment_'~i] and link_url ) %}
-								<div class="four-column-symbol-container">
-									<div class="four-column-symbol">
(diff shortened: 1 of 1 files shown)


includes/blocks/static_four_column.twig
```diff
@@ -7,39 +7,41 @@
...
-								<div class="four-column-symbol-container">
-									<div class="four-column-symbol">
+						{% if fields['title_'~i] %}
+							<div class="col-md-6 col-lg four-column-wrap">
+								{% set link_url = fields['link_url_'~i] %}
+								{% if ( fields['attachment_'~i] and link_url ) %}
+									<div class="four-column-symbol-container">
+										<div class="four-column-symbol">
+											<a href="{{ link_url|e('esc_url') }}">
+												<img src="{{ fields['attachment_'~i] }}" alt="">
```
angtheod: Don't we want an alt text here and in line 23?

comzeradd:
@angtheod I amended the commit. I didn't touch the code of the block besides wrapping it up in an if statement, but you are right. These alt tags shouldn't be empty, so I added the element title there.

PR #247, opened 2018-03-27
PLANET-1805 Hiding the carousel-split from the available page elements for now

koyan:


angtheod:
Is the idea to 'disable' this block for now? If so, we could comment out the line that adds it in the plugin's loader method here https://github.com/greenpeace/planet4-plugin-blocks/blob/develop/planet4-blocks.php#L66 to skip running its code completely.

koyan:
@angtheod Done as you suggested

PR #248, opened 2018-03-27
PLANET-1981-Added default option in select issue dropdown

sagarsdeshmukh:


angtheod:
Need a confirmation on expected behavior if user leaves the default behavior (not select any issue at all).

koyan:
@angtheod @sagarsdeshmukh If the user leaves the default option (aka, not select an issue), then no background image should be shown. Any field that is not also selected is to be left empty

kirdia:
@sagarsdeshmukh Some part of the template/html will be rendered with this approach, even if no issue is selected. Is this the wanted behavior?

sagarsdeshmukh:
@kirdia Yes, please check the updated discussion [here](https://jira.greenpeace.org/browse/PLANET-1981?focusedCommentId=62788&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-62788).

PR #253, opened 2018-03-30
PLANET-2013 Added parameter for ignoring categories filtering

koyan:
so that pagetypes can be set as the only criteria
Also, made pagetype appear even if there is no tag

### DIFF
--- a/classes/controller/blocks/class-articles-controller.php
+++ b/classes/controller/blocks/class-articles-controller.php
@@ -34,6 +34,14 @@ if ( ! class_exists( 'Articles_Controller' ) ) {
 
 			// Construct a checkbox for each p4-page-type.
 			if ( ! empty( $planet4_article_type_terms ) ) {
+				$checkboxes [] = [
+					'attr'        => 'ignore_categories',
+					'label'       => 'Ignore Categories',
+					'description' => 'Ignore categories when filtering posts to populate the content of this block',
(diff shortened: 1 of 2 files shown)


classes/controller/blocks/class-articles-controller.php
```diff
@@ -34,6 +34,14 @@ public function prepare_fields() {
 
 			// Construct a checkbox for each p4-page-type.
 			if ( ! empty( $planet4_article_type_terms ) ) {
+				$checkboxes [] = [
```
kirdia: Minor thing, extra space between $checkboxes and []

classes/controller/blocks/class-articles-controller.php
```diff
@@ -177,9 +185,12 @@ public function prepare_template( $fields, $content, $shortcode_tag ) : string {
 			}
 			$fields['read_more_link'] = $read_more_link;
 
+
 			$category_ids = '';
-			if ( $category_id_array ) {
-				$category_ids = implode( ',', $category_id_array );
+			if ( ! $fields['ignore_categories'] ) {
```
kirdia: If we are going to ignore categories based on the param, then it is better to remove it from args array (for the posts query) and add it only we do need categories. Like tags for example
`			if ( $category_ids ) {
				$args['category'] = '( ' . $category_ids . ' )';
			}
`

PR #262, opened 2018-04-13
PLANET-1943 Initial version of submenu block

kirdia:
[Issue PLANET-1943](https://jira.greenpeace.org/browse/PLANET-1943)


classes/controller/blocks/class-submenu-controller.php
```diff
@@ -0,0 +1,334 @@
...
+			$link1    = $attributes['link1'] ?? 'false';
+			$link2    = $attributes['link2'] ?? 'false';
+
+			$menu = $this->parse_post_content( $content, [ $heading1, $heading2 ], [ $link1, $link2 ] );
+
+			wp_enqueue_script( 'submenu', P4BKS_ADMIN_DIR . 'js/submenu.js' );
+			wp_localize_script(
+				'submenu',
+				'submenu',
+				wp_json_encode( $menu )
```
angtheod: 3rd arg of wp_localize_script should be array but wp_json_encode returns false/string.
So, we should put it in an array like
`[ wp_json_encode( $menu ) ]`

classes/controller/blocks/class-submenu-controller.php
```diff
@@ -0,0 +1,334 @@
...
+				}
+			}
+
+			return $menu;
+		}
+
+		/**
+		 * Create a std object representing a node/heading.
+		 *
+		 * @param \DOMElement $node
```
angtheod: missing doc comments

classes/controller/blocks/class-submenu-controller.php
```diff
@@ -0,0 +1,334 @@
...
+			$heading              = intval( $heading );
+			if ( ! empty( $heading ) &&
+				in_array( $heading, $valid_heading_values, true ) ) {
+				return true;
+			}
+
+			return false;
+		}
+	}
+}
```
angtheod: please add a new line at the end of the file

admin/templates/submenu.tpl.php
```diff
@@ -0,0 +1,49 @@
+<script type="text/html" id="tmpl-shortcode-ui-field-p4-radio">
```
angtheod: Does this file need to be a .php? Codesniffer complains that it did not find any php code or php short open tags

classes/controller/blocks/class-submenu-controller.php
```diff
@@ -0,0 +1,334 @@
...
+						],
+						[
+							'value' => '2',
+							'label' => __( 'Short full-width', 'planet4-blocks' ),
+							'desc'  => 'Use: on long pages (more than 5 screens) when list items are short (up to 10 words)<br>Max items<br>recommended:9<br>Example',
+						],
+						[
+							'value' => '3',
+							'label' => __( 'Short sidebar', 'planet4-blocks' ),
+							'desc'  => 'Use: on long pages (more than 5 screens) when list items are short (up to 10 words)<br>Max items<br>recommended:9<br>Example',
```
angtheod: Perhaps different descriptions for the 3 radio buttons have not been provided by the design, but just mentioning this

angtheod:
Overall looks good, just some minor comments.

classes/controller/blocks/class-submenu-controller.php
```diff
@@ -0,0 +1,334 @@
...
+		 *
+		 * @param string $template Template name.
+		 *
+		 * @return bool|string
+		 */
+		private function get_template( $template ) {
+
+			$template = P4BKS_PLUGIN_DIR . '/admin/templates/' . $template . '.tpl.php';
+			if ( file_exists( $template ) ) {
+				$contents = file_get_contents( $template );
```
angtheod: if possible replace file_get_contents with wp_remote_get as WP CS suggest

classes/controller/blocks/class-submenu-controller.php
```diff
@@ -0,0 +1,334 @@
...
+				'listItemImage' => '<img src="' . esc_url( plugins_url() . '/planet4-plugin-blocks/admin/images/submenu.png' ) . '" />',
+				'attrs'         => $fields,
+				'post_type'     => P4BKS_ALLOWED_PAGETYPE,
+			];
+
+			shortcode_ui_register_for_shortcode( 'shortcake_' . self::BLOCK_NAME, $shortcode_ui_args );
+		}
+
+		/**
+		 * Callback for content four column shortcode.
```
sagarsdeshmukh: Doc block description needs to change.

admin/templates/submenu.tpl.php
kirdia: Actually no, it does not have to be a php file. I just followed shortcake's convention, which names underscore templates with the extension .tpl.php. This file contains html with underscore commands and all these templates are appended at the end of the dom, to be used from backbone views objects.

classes/controller/blocks/class-submenu-controller.php
kirdia: Codesniffer locally suggests using WP_Filesystem, but for some reason which I don't remember I did not use it. I'll try it again.
No need to use wp_remote_get, it is a local file.

PR #282, opened 2018-05-24
PLANET-2219 Initial version of cookies block

kirdia:
[Issue 2219](https://jira.greenpeace.org/browse/PLANET-2219)
Cookies block will set 'greenpeace' cookie according to which type of cookies the user will select.


comzeradd:
JS code looks good too :+1:

PR #283, opened 2018-06-01
PLANET-2307 Make the campaign image use the correct srcset in split-two-column

koyan:


classes/controller/blocks/class-splittwocolumns-controller.php
```diff
@@ -253,9 +253,7 @@ public function prepare_template( $fields, $content, $shortcode_tag ) : string {
 				],
 				'campaign' => [
 					'image'       => wp_get_attachment_url( $campaign_image_id ),
-					'srcset'      => wp_calculate_image_srcset( [ '1118', '746' ],
-										wp_get_attachment_image_src( $campaign_image_id, 'full' )[0],
-										wp_get_attachment_metadata( $campaign_image_id ) ),
+					'srcset'      => wp_get_attachment_image_srcset( $campaign_image_id ),
```
kirdia: Is this generates srcset for medium & thumbnail sizes only?
'medium' is 300px x 300px. I think it is too small for this block.
`wp_get_attachment_image_srcset( $campaign_image_id, 'large' )` will be more appropriate. It will generate srcset for sizes 'large' and below.

PR #287, opened 2018-06-12
PLANET-2305 Fix php warning in preview_iframe function

kirdia:


classes/controller/blocks/class-controller.php
```diff
@@ -126,17 +126,16 @@ public function shortcode_ui_register_shortcodes() {
...
-					]
-				),
+			$preview_args   = [
+				'_tag'     => $shortcode_tag,
+				'_content' => $content,
+				'_post_id' => get_the_ID(),
+				'_nonce'   => wp_create_nonce( 'p4bks_preview_render_' . static::BLOCK_NAME ),
+				'action'   => 'p4bks_preview_render_' . static::BLOCK_NAME,
+			];
+			$request_params = ! is_null( $fields ) && ! empty( $fields ) ? array_merge( $fields, $preview_args ) : $preview_args;
```
angtheod: Looks good that way, but just an idea, If we put $preview_args as 1st arg in array_merge() then I think it won't matter if $fields is  empty, since array_merge can work with 1 array as well, right?

PR #302, opened 2018-07-10
PLANET-2149 add enform block

angtheod:
Allow registration of a shortcake block from outside of this plugin. The external blocks are registered after all plugins are loaded. Renamed View argument to make more sense.

related to this [PR](https://github.com/greenpeace/planet4-plugin-engagingnetworks/pull/21)

### DIFF
--- a/classes/class-loader.php
+++ b/classes/class-loader.php
@@ -15,6 +15,10 @@ if ( ! class_exists( 'Loader' ) ) {
 		private static $instance;
 		/** @var array $services */
 		private $services;
+		/** @var array $external_services */
+		private $external_services = [];
+		/** @var Views\View $view */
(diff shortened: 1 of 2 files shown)


classes/class-loader.php
```diff
@@ -45,26 +49,49 @@ public static function get_instance( $services = array(), $view_class ) : Loader
 		 * @param string $view_class The View class name.
 		 */
 		private function __construct( $services = array(), $view_class ) {
+			$this->load_services( $services, $view_class );
+			$this->check_requirements();
+			add_action( 'plugins_loaded', array( $this, 'load_i18n' ) );
```
kirdia: @angtheod Can you change the array to short syntax for consistency?

PR #306, opened 2018-07-13
PLANET-2356 - Articles block: Autoscale featured image thumbnail

comzeradd:


angtheod:
@comzeradd is this going into 1.23?

comzeradd:
Yes, good catch. I just added the tag on the ticket.

PR #314, opened 2018-08-03
Clean up css code

comzeradd:
* Rename scss files to match block names
* Use rem for font-size
* Use px for line-height and margin
* Clean up comments
* Refactor inconsistent media queries
* Add cookies block


assets/scss/blocks/_cookies.scss
```diff
@@ -0,0 +1,115 @@
+.cookie-block {
+  z-index: 100;
+  display: none;
+  position: fixed;
+  bottom: 0;
+  left: 0;
+  width: 100%;
+  padding: 10px 0;
+  background: $cookie-bkg url("images/cookie-right-gradient.png") no-repeat right;
```
kirdia: @comzeradd the cookies block images are missing from the repo (images/cookie-right-gradient-[sm|md].png)

comzeradd: Good catch. I amended the commit to move these images here.

PR #319, opened 2018-08-08
PLANET-2515 unit test articles block

angtheod:
Setup files and packages for WP plugin Unit tests having automation in mind. Created first Unit test for Articles block.


angtheod:
@comzeradd Nikos please check the .travis.yml file.
This was automatically editted by the wp scaffold command. Do we need to keep the lines that were there before?

comzeradd:
> Do we need to keep the lines that were there before?

Ideally yes, in order to run the linting tests for the frontend code.

We need this on `before_script`:
```
- npm install -g stylelint eslint
```

and these two lines on `script`:
```
- stylelint 'assets/scss/**/*.scss' 'assets/css/*.css'
- eslint 'assets/js/**/*.js'
```

angtheod:
:+1: Yes, I wasn't aware of that flag. Will change that as well.

angtheod:
I had to refactor the prepare_template method of the Articles controller class.
Needed to separate logic for retrieving the required data into another method which is now testable.

@27Bslash6 @comzeradd Where should the lint lines go inside the circle.yml file?

comzeradd:
> @27Bslash6 @comzeradd Where should the lint lines go inside the circle.yml file?

The install part should probably be inside the `dependencies` block and the actual tests commands inside `test: pre:`. I'm not about the output format though. Does circleci expects only junit? /cc @27Bslash6

classes/controller/blocks/class-articles-controller.php
```diff
@@ -275,20 +289,22 @@ public function prepare_template( $fields, $content, $shortcode_tag ) : string {
 					$page_type_data = get_the_terms( $recent['ID'], 'p4-page-type' );
 					$page_type      = '';
 
-					if ( $page_type_data ) {
+					if ( $page_type_data && ! is_wp_error( $page_type_data ) ) {
```
angtheod: This correction came as a result of actually running the newly created Unit tests.

classes/controller/blocks/class-articles-controller.php
```diff
@@ -275,20 +289,22 @@ public function prepare_template( $fields, $content, $shortcode_tag ) : string {
...
 					}
 
 					$recent['page_type'] = $page_type;
 					$recent['permalink'] = get_permalink( $recent['ID'] );
 
-					$recent['filter_url'] = add_query_arg( [
-							's'                        => ' ',
-							'orderby'                  => 'relevant',
-							'f[ptype]['.$page_type.']' => $page_type_id,
+					if ( isset( $page_type_id ) ) {
```
angtheod: Same here, this extra check came as a result of actually running the newly created Unit tests.

angtheod:
Must be merged together with #331

27Bslash6:
Yes CircleCI wants junit.xml output format for storing and parsing tests results.

As far as the CI test scaffold output goes, I'm not experienced with the machine executor - I always use the docker executor for speed reasons. Furthermore, I don't know whether the functions initialising environment will work as intended. I think we should suck it and see, so approving the changes.

PR #323, opened 2018-08-10
PLANET-2412 Add sixth scenario in articles block.

kirdia:
Select posts based on page's tags.
When no page types, tags or specific posts are selected it will select posts based on the page's tags that the shortcode is embedded into.

Fixes explore pages.
https://jira.greenpeace.org/browse/PLANET-2412?focusedCommentId=66757&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-66757

### DIFF
--- a/classes/controller/blocks/class-articles-controller.php
+++ b/classes/controller/blocks/class-articles-controller.php
@@ -635,21 +635,16 @@ if ( ! class_exists( 'Articles_Controller' ) ) {
 
 			// For posts and pages, display related articles based on current post/page tags.
 			$current_post_type = get_post_type();
+			$read_more_filter  = '';
+
 			if ( 'post' === $current_post_type || 'page' === $current_post_type ) {
 				if ( $post_tags ) {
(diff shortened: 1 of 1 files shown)


classes/controller/blocks/class-articles-controller.php
```diff
@@ -610,6 +613,52 @@ private function filter_posts_for_tag_page( &$fields ) {
...
+					}
+					$args['tag__in'] = $tag_id_array;
+				}
+			}
+
+			$read_more_filter = '';
+			// For normal page and post.
+			if ( $post_tags ) {
+				foreach ( $post_tags as $tag ) {
+					$read_more_filter .= '&f[tag][' . $tag->name . ']=' . $tag->term_id;
```
angtheod: Can't this go into the foreach in line 641?

kirdia: Yeap, good catch. I moved it in the first loop :+1:

PR #334, opened 2018-08-22
Fix composer autoload files.

angtheod:
Run composer update --no-dev and composer dump-autoload --optimize in order to fix composer's autoload files because last time I had forgotted to add --no-dev and therefore the autoload files tried to load also the dependencies files (which were not found since they were being git-ignored).

(Btw, re-added the WP Coding Standards packages in require-dev)


angtheod:
Removed require-dev dependencies completely

Issue #335, opened 2018-08-23
Update CircleCI configuration to version 2.0


27Bslash6:
See: https://circleci.com/gh/greenpeace/planet4-plugin-blocks/6

`This project is currently running on CircleCI 1.0 which will no longer be supported after August 31, 2018. Please start migrating this project to CircleCI 2.0`

@angtheod I guess that when `wp scaffold` was run, it was with a version of wp-cli from before this PR was merged: https://github.com/wp-cli/scaffold-command/pull/115

Perhaps try rebuilding the configuration with a recent version of wp-cli

Builds on Circle v1.0 will stop working on August 31

angtheod:
Oh, didn't know that, sorry!

PR #337, opened 2018-08-23
Updated circle ci config file to 2.0 by updating wp-cli to 2.0.0

angtheod:
Issue https://github.com/greenpeace/planet4-plugin-blocks/issues/335

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -10,7 +10,7 @@ version: 2
 
 job-references:
   mysql_image: &mysql_image
-    circleci/mysql:5.6
+    circleci/mysql:5.7
 
   setup_environment: &setup_environment
     name: "Setup Environment Variables"
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -35,13 +35,18 @@ job-references:
       - run: *setup_environment
       - run: *install_dependencies
       - run:
-          name: "Run Tests"
+          name: "phpcs"
           command: |
-            mkdir -p /tmp/test-results/phpcs /tmp/test-results/phpunit /tmp/test-results/phpunit-multisite
-            composer global require "phpunit/phpunit=5.7.*"
+            mkdir -p /tmp/test-results/phpcs
             composer global require wp-coding-standards/wpcs
             phpcs --config-set installed_paths $HOME/.composer/vendor/wp-coding-standards/wpcs
             phpcs --report=junit --report-file=/tmp/test-results/phpcs/$CIRCLE_STAGE.xml
+      - run:
+          name: "phpunit"
+          when: always
+          command: |
+            mkdir -p /tmp/test-results/phpunit /tmp/test-results/phpunit-multisite
+            composer global require "phpunit/phpunit=5.7.*"
             rm -rf $WP_TESTS_DIR $WP_CORE_DIR
             bash bin/install-wp-tests.sh wordpress_test root '' 127.0.0.1 latest
             phpunit --log-junit /tmp/test-results/phpunit/$CIRCLE_STAGE.xml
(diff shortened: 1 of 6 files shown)


.circleci/config.yml
```diff
@@ -0,0 +1,95 @@
...
+    command: |
+      sudo apt-get update && sudo apt-get install subversion
+      sudo docker-php-ext-install mysqli
+      sudo sh -c "printf '\ndeb http://ftp.us.debian.org/debian sid main\n' >> /etc/apt/sources.list"
+      sudo apt-get update && sudo apt-get install mysql-client-5.7
+
+  php_job: &php_job
+    environment:
+      - WP_TESTS_DIR: "/tmp/wordpress-tests-lib"
+      - WP_CORE_DIR: "/tmp/wordpress/"
```
angtheod: @27Bslash6 is it ok if it downloads the wordpress_tests suite inside /tmp?

.circleci/config.yml
```diff
@@ -0,0 +1,95 @@
...
+    steps:
+      - checkout
+      - run: *setup_environment
+      - run: *install_dependencies
+      - run:
+          name: "Run Tests"
+          command: |
+            composer global require "phpunit/phpunit=5.7.*"
```
angtheod: @27Bslash6 Should I change that to 6.5.* ? Or it does not have an effect either way?

.circleci/config.yml
```diff
@@ -0,0 +1,95 @@
...
+            composer global require wp-coding-standards/wpcs
+            phpcs --config-set installed_paths $HOME/.composer/vendor/wp-coding-standards/wpcs
+            phpcs
+            rm -rf $WP_TESTS_DIR $WP_CORE_DIR
+            bash bin/install-wp-tests.sh wordpress_test root '' 127.0.0.1 latest
+            phpunit
+            WP_MULTISITE=1 phpunit
+
+jobs:
+  php56-build:
```
angtheod: @27Bslash6 should I remove the php5.6 job?

27Bslash6: Yes we'll never need the 5.6 tests

.circleci/config.yml
27Bslash6: Yes, no worries. You can manipulate the file system any way you require.

.circleci/config.yml
27Bslash6: I don't know whether the version would have any effect.  If this is boilerplate from `wp scaffold`, we can start with 5.7.* and then update to 6.5 in a future commit.

.phpcs.xml.dist
```diff
@@ -0,0 +1,48 @@
...
+			<!-- Value: replace the text domain used. -->
+			<property name="text_domain" type="array" value="my-plugin"/>
+		</properties>
+	</rule>
+	<rule ref="WordPress.WhiteSpace.ControlStructureSpacing">
+		<properties>
+			<property name="blank_line_check" value="true"/>
+		</properties>
+	</rule>
+</ruleset>
```
27Bslash6: This file fails to load into phpcs: https://circleci.com/gh/greenpeace/planet4-plugin-blocks/14
```
ERROR: Ruleset /home/circleci/project/.phpcs.xml.dist is not valid

Run "phpcs --help" for usage information

Exited with code 3
```
I've tested this install with the default file https://raw.githubusercontent.com/WordPress-Coding-Standards/WordPress-Coding-Standards/develop/phpcs.xml.dist.sample and phpcs successfully parses the config file, but the style / standards may not be what you're used to.

Can you have a look at this file and see if it's more appropriate, or fix the existing so at least the tests run, if not succeed?

27Bslash6:
Alright this now executes phpcs and phpunit tests, and stores the results in a format that Circle can summarise. 

Next steps:
- #339 solve the phpcs errors
- increase unit test coverage
- automate the deployment of new releases

PR #364, opened 2018-10-11
PLANET-2751 Fix posts limit in new covers block

kirdia:
Fix posts limit for custom ids case in new covers block.
Change posts ordering in content covers case to date descending.


angtheod:
It's not in the PR but this return comment needs to be removed [line 535](https://github.com/greenpeace/planet4-plugin-blocks/pull/364/files#diff-bfe657539f6bdab15f3a61d5568a6155R535)

btw, nice trick with the dataProvider

PR #375, opened 2018-10-23
Use Page Type slug in articles block.

pablocubico:
As I changed Page Type links in the master theme, I'm doing the same here for consistency.

Ref. issue: https://jira.greenpeace.org/browse/PLANET-2614

More details on: https://github.com/greenpeace/planet4-master-theme/pull/587

### DIFF
--- a/classes/controller/blocks/class-articles-controller.php
+++ b/classes/controller/blocks/class-articles-controller.php
@@ -294,15 +294,9 @@ if ( ! class_exists( 'Articles_Controller' ) ) {
 					}
 
 					$recent['page_type'] = $page_type;
-					$recent['permalink'] = get_permalink( $recent['ID'] );
+					$recent['page_type_id'] = $page_type_id;
 
-					if ( isset( $page_type_id ) ) {
-						$recent['filter_url'] = add_query_arg( [
-							's'                            => ' ',
-							'orderby'                      => 'relevant',
-							'f[ptype][' . $page_type . ']' => $page_type_id,
-						], get_home_url() );
-					}
+					$recent['permalink'] = get_permalink( $recent['ID'] );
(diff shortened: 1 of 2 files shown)


includes/blocks/articles.twig
```diff
@@ -48,7 +48,7 @@
 									{% if ( value.tags or value.page_type ) %}
 										<div class="article-list-item-tags top-page-tags">
 											{% if ( value.page_type ) %}
-												<a class="tag-item tag-item--main page-type" href="{{ value.filter_url }}">{{ value.page_type|e('wp_kses_post')|raw }}</a>
+												<a class="tag-item tag-item--main page-type" href="{{ fn('rtrim', fn('get_home_url'), '/') ~ '/page_type/' ~ value.page_type_slug }}">{{ value.page_type|e('wp_kses_post')|raw }}</a>
```
kirdia: Same here as master-theme.
We can use [get_term_link](https://developer.wordpress.org/reference/functions/get_term_link/).

classes/controller/blocks/class-articles-controller.php
```diff
@@ -289,11 +289,14 @@ private function populate_post_items( $posts ) {
...
+						$page_type_id   = $page_type_data[0]->term_id;
+						$page_type_slug = $page_type_data[0]->slug;
 					}
 
 					$recent['page_type'] = $page_type;
+					$recent['page_type_slug'] = $page_type_slug;
+
 					$recent['permalink'] = get_permalink( $recent['ID'] );
 
 					if ( isset( $page_type_id ) ) {
```
kirdia: Also, please remove that part, the filter_url becomes obsolete, so no need to populate it.

PR #376, opened 2018-10-24
Feature/2472 video block

sagarsdeshmukh:
Task:
- Change block name from "Youtube video block" to "Video block".
- Expand existing Youtube video block functionality to serve the GP Media library video.

GP media library video links -
https://media.greenpeace.org/GPIDoc/GPI/Media/Video/WebHigh/9/d/8/7/GP0STS2AU.mp4
https://media.greenpeace.org/GPIDoc/GPI/Media/Video/WebHigh/c/e/9/e/GP0STSH47.mp4
https://media.greenpeace.org/GPIDoc/GPI/Media/Video/WebHigh/f/d/0/f/GP0STS7M1.mp4

### DIFF
--- a/classes/controller/blocks/class-mediavideo-controller.php
+++ b/classes/controller/blocks/class-mediavideo-controller.php
@@ -22,24 +22,24 @@ if ( ! class_exists( 'MediaVideo_Controller' ) ) {
 		 * @since 0.1.0
 		 */
 		public function prepare_fields() {
-			$fields = array(
-				array(
+			$fields = [
+				[
 					'label' => __( 'Video Title', 'planet4-blocks-backend' ),
 					'attr'  => 'video_title',
 					'type'  => 'text',
-					'meta'  => array(
+					'meta'  => [
 						'placeholder' => __( 'Enter video title', 'planet4-blocks-backend' ),
-					),
-				),
-				array(
+					],
+				],
+				[
 					'label' => __( 'Video URL/Youtube ID', 'planet4-blocks-backend' ),
 					'attr'  => 'youtube_id',
 					'type'  => 'text',
-					'meta'  => array(
+					'meta'  => [
 						'placeholder' => __( 'Enter video URL or youtube video id', 'planet4-blocks-backend' ),
-					),
-				),
-				array(
+					],
+				],
+				[
 					'label'       => __( 'Video thumbnail image [Optional]', 'planet4-blocks-backend' ),
 					'attr'        => 'video_thumbnail_img',
 					'type'        => 'attachment',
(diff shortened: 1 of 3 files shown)


classes/controller/blocks/class-mediavideo-controller.php
```diff
@@ -63,6 +72,29 @@ public function prepare_fields() {
...
+					if ( isset( $matches[5] ) && $matches[5] ) {
+						// Extract youtube video ID and use with youtube embed url format.
+						$fields['youtube_id']       = $matches[5];
+						$fields['is_youtube_video'] = true;
+					}
+				} else {
+					// Case 3 : Video URL other than Youtube (GP media library video etc).
+					$fields['is_youtube_video'] = false;
+					if ( $fields['video_thumbnail_img'] ) {
+						$fields['video_thumbnail_src'] = wp_get_attachment_image_src( $fields['video_thumbnail_img'], 'retina-large' );
```
kirdia: We can use a smaller thumbnail size, since the 'video-section' div is constrained by css. I think 'large' would be enough.

includes/blocks/media_video.twig
```diff
@@ -6,9 +6,16 @@
...
-					<iframe id="ytplayer" type="text/html" width="100%" height="100%"
-					src="https://www.youtube.com/embed/{{ fields.youtube_id }}?autoplay=0&rel=0"
-					frameborder="0" allowfullscreen> </iframe>
+					{% if ( fields.is_youtube_video ) %}
+						<iframe id="ytplayer" type="text/html" width="100%" height="100%"
+						src="https://www.youtube.com/embed/{{ fields.youtube_id }}?autoplay=0&rel=0"
+						frameborder="0" allowfullscreen>
+						</iframe>
+					{% else %}
+						<video {% if ( fields.video_thumbnail_src ) %}poster="{{ fields.video_thumbnail_src.0 }}"{% endif %} controls="controls" width="100%" height="100%">
```
kirdia: Here for presentation can you add a css rule so that poster will use all of the container's size?
For example:
`object-fit: cover;`

includes/blocks/media_video.twig
```diff
@@ -6,9 +6,16 @@
...
+							<source src="{{ fields.youtube_id }}" type="video/mp4" />
```
kirdia: Also here, I don't know if it makes any difference for the video tag, but the filetype will always be mp4?

classes/controller/blocks/class-mediavideo-controller.php
sagarsdeshmukh: Yes, I will change it to large. Also I think I need to change the label of image as "Video poster image [Optional]" instead of "Video thumbnail image [Optional]". your comment?

PR #381, opened 2018-10-31
PLANET-2731 Fix phpcs errors

kirdia:
Fixed errors/warning using the latest wordpress sniffer rules (1.1.0) [worpress-coding-standards](https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards)
Remove old phpcs file.
Added rules configuration in phpcs file.
Adhering to rulesets:
- Wordpress-Core
- Wordpress-Docs
- Wordpress-Extra


composer.json
```diff
@@ -9,7 +9,10 @@
       "email": "planet4-group@greenpeace.org"
     }
   ],
-  "require-dev": {},
+  "require-dev": {
+    "squizlabs/php_codesniffer": "^3.3.2",
+    "wp-coding-standards/wpcs": "^1.1.0"
```
comzeradd: I think we also need `dealerdirect/phpcodesniffer-composer-installer`

kirdia: I have tried using it, but it adds entries in composer's autoload files.
We use the composer's autoload files for autoloading the repo's classes, so I think it will be a problem in production.

comzeradd: We already have it on master-theme. But on [deployment](https://github.com/greenpeace/planet4-builder/blob/develop/src/build/Dockerfile.in#L22) we run composer with `--no-dev` flag.

kirdia: The plugin won't exist in production, but the autoload files will still contain the entries for the plugin. 
Still it won't be a problem now that I'm thinking of it, because we will not be using the classes of the plugin in our code, but I think we should keep the autoload files as clean as possible.

Another solution would be to use that when we want to add/edit entries for our classes, which won't add entries for the dev dependencies.
`composer dumpautoload --optimize --no-dev`

But if we want to use it locally then we have to revert changes back locally with that
`composer dumpautoload --optimize`

It is a bit contradicting, what do you think?

comzeradd: I have forgotten that we have the vendor folder checked in to the repo. So, let's keep `phpcodesniffer-composer-installer` out for now.

Feel free to merge :+1:

PR #384, opened 2018-11-05
PLANET - no issue: Fix TinyMCE disappearing from Post edit

pablocubico:
Editor disappeared from Post Edit page. Looks like the `wp_filesystem` call is causing the problem.

I just restored the original version, feel free to change it if you find a better method.

Screenshot:
image


comzeradd:
I'm merging this just to unblock the UAT. But we should figure out how to fix it and also make `phpcs` happy.

PR #396, opened 2018-11-14
PLANET-2395 Fix tasks step numbers hover

kirdia:


assets/js/partials/can_do_steps.js
```diff
@@ -1,8 +1,8 @@
-$(document).ready(function() {
+$(document).ready(function () {
   'use strict';
 
-  $('.can-do-steps .col').hover(function() {
+  $('.can-do-steps .col').hover(function () {
     const step = $(this).data('id');
-    $('#step-' + step).toggleClass('active');
+    $(this).find('#step-' + step).toggleClass('active');
```
pablocubico: We shouldn't need to use `find` for an element ID, as IDs are supposed to be unique. I guess there are other `#step-...` elements in the page rendered somewhere else. If that's the case, I suggest either using a "uniquer" ID other than "step-..." to remove the `find()` call or switching it to a `.step-...` class name and keep using `find()`.

kirdia: You are right. Removed completely the step-ids as it were used only for this purpose.

kirdia:
Added also a small fix for mobile accordion divs (open the right target when multiple tasks blocks exist)

PR #398, opened 2018-11-16
PLANET-2873 fix read more link

angtheod:
Remove uneccessary encode of read_mode_link url.


admin/js/blocks-ui.js
```diff
@@ -419,7 +419,7 @@ p4_blocks = {
       var view = p4_blocks.find_view(collection, 'read_more_link');
       if ('undefined' !== view) {
         if ('undefined' !== changed.value) {
-          var url = encodeURI(changed.value);
+          var url = changed.value;
```
kirdia: I haven't tested it locally, but I think it will break the block when user enters urls containing some special chars.
I could not find the related issue, but this was introduced here.
https://github.com/greenpeace/planet4-plugin-blocks/commit/b8f8c9fc188a3bef01c9534bbd572a753458124b#diff-34cab90e5d8c7b772a6b3152cee44df9

The thing is that shortcake by default does not escape square brackets ( '[', ']' ), so if the user enters a bracket in the input field, it results in a broken shortcode.

angtheod: Hmm good point, so I guess we can encode it only if it is not already encoded or just encode '[', ']'

angtheod:
@comzeradd @kirdia @sagarsdeshmukh Please, check again this one, so that it moves to UAT if it gets merged.

PR #401, opened 2018-11-19
PLANET-2559 articles load more

angtheod:
Implement Load more functionality for the Articles block. Separate template of each article's post from the full articles template. Get page type, tag and author data from the update P4_Post (TimberPost) and P4_User (TimberUser) objects. This way we can use Timber's functionality easier. Removed no longer neccessary code and markup.

Note:
To make use of the JS load_more functionality for another block, you need to add the following to your block's button:
`load_more` (as a class)
`data-content="<selector>"` (as an attribute)
e.g. if <selector> is `.article-list-section` then the new content will be appended there


angtheod:
Related with https://github.com/greenpeace/planet4-master-theme/pull/629

PR #404, opened 2018-11-20
Add margin to `campaign-card-column` selector.

Wakkatela:


Wakkatela:
We should not push .css or css.map neither. The package was created when running *npm* to install plugins.

Should I remove all `.css`, `.map` and `.json` files? Or only the `.json`?

comzeradd:
We do want minified minified css and map. At least until we figure out a better way to generate them on deploy.

assets/scss/blocks/_campaign_thumbnail.scss
```diff
@@ -32,6 +32,7 @@
 
     .campaign-card-column {
       padding: 8px inherit;
+      margin-bottom: 32px;
```
comzeradd: The padding attribute on top of that has invalid syntax. I think if you just change it to `padding: 8px` there will be no need for adding margin.

Wakkatela: Ok, I'll do that, I just felt like not changing too much of what's already there =D

Wakkatela: I did the changes, but, in these cases I prefer to use Margins (we want the element to be apart from others, not from its box limit). We could set those rules in the refactoring.

comzeradd: Yes, you are right. I haven't tried it, but I guess `margin: 8px` should have the same visual result on this case.

comzeradd:
Btw, the reason we don't keep `package-lock.json` is because we already have `yarn.lock` there. If we think that it's easier to just use npm instead of yarn, we can revisit that in the future.

PR #408, opened 2018-11-22
PLANET-2932 Load all post on articles block load more

sagarsdeshmukh:
Note: Here we are fetching all posts from DB, If there are thousands of posts then it will impact on page load performance.

### DIFF
--- a/classes/controller/blocks/class-articles-controller.php
+++ b/classes/controller/blocks/class-articles-controller.php
@@ -245,6 +245,10 @@ For good user experience, please include at least three articles so that spacing
 			} else {
 				$args = $this->filter_posts_by_pages_tags( $fields );
 			}
+
+			// Get all posts.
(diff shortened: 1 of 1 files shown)


classes/controller/blocks/class-articles-controller.php
```diff
@@ -245,6 +245,10 @@ public function prepare_data( $fields, $content = '', $shortcode_tag = 'shortcak
 			} else {
 				$args = $this->filter_posts_by_pages_tags( $fields );
 			}
+
+			// Get all posts.
+			$args['numberposts'] = -1;
```
angtheod: Would be better to set a high enough number instead of -1 (e.g. 100)

koyan: I thought about this also Angelos. Want to do a PR for it?

PR #413, opened 2018-11-28
PLANET-2933: Articles block: Load More button is wrong colour - Button fix

pablocubico:
The `<a>` tag was causing a request to be fired.

Ref: https://jira.greenpeace.org/browse/PLANET-2933


comzeradd:
The only problem I see with this approach is that the button won't work if js is disabled, right?

pablocubico:
Hey @comzeradd ! The `e.preventDefault()` call doesn't prevent the request from firing in `mousedown` events (as far as I recall). In fact, I think we could get rid of `e.preventDefault()` 🤔 , I tend to be over-defensive...

includes/blocks/articles.twig
```diff
@@ -27,16 +27,15 @@
...
+								<button class="btn btn-secondary btn-block load-more"
 								   data-content=".article-list-section"
 								   data-page="1"
 								   data-total_pages="{{ total_pages }}"
 								   data-article_count="{{ fields.article_count }}"
 										{% for key,value in dataset %}
 											data-{{ value }}
 										{% endfor %}
-								>{{ fields.read_more_text }}</a>
+								>{{ fields.read_more_text }}<buttona>
```
kirdia: needs a small fix in the ending tag here

pablocubico: Good catch! Fixing

comzeradd:
Can you squash the commits? That's a good example of not making [typo-fixing commits](https://planet4.greenpeace.org/handbook/dev-git-best-practices/#amend) :D 

Or we could also squash it through Github on merge.

pablocubico:
@comzeradd squashed!

PR #420, opened 2018-12-06
PLANET-3014: Initial implementation of Counter block

tamlyn:
This is still a work in progress but opening this PR for some feedback. Specifically:

- Is [PLANET-2609](https://jira.greenpeace.org/browse/PLANET-2609) the right issue to link this to?
- We have rough visuals for the six campaign styles but what should the base (un)style be?
- How do blocks size themselves and how does this change on different devices? 
- Are there some general responsive principles common to all blocks?
- Should `style.css` and `style.css.map` be included in the PR?

Any other feedback on code style or structure welcomed.


koyan:
Thanks for that!

> Is PLANET-2609 the right issue to link this to?
The original ticket was for the design team. I have created a ticket so that you can tag that one for the development: PLANET-3014

> We have rough visuals for the six campaign styles but what should the base (un)style be?
We are checking with our design team and will get back to you about it. Since the block will be available thoughout the whole platform, and an editor will be able to put in non campaign pages, we need to make sure that it looks good on the default master theme

I will let the other questions be answered by Nikos ( @comzeradd ) and the design team. 
A couple of notes: You can see that the ci/circleci tests have failed. (Feel free to ignore the codacy ones. They should not be there to start with). Please check that feedback (mostly styling issues). 

@kirdia / @sagarsdeshmukh / @angtheod will also look at the php part of it as well.

kirdia:
> Should style.css and style.css.map be included in the PR?

Yes, we currently check in changes in style.css/style.css.map

 > Any other feedback on code style or structure welcomed.
There are some code style validation errors. I don't know if you can see the ci builds.
But you can use composer to install the dev dependencies and then run `phpcs` to see the validation errors locally.

classes/controller/blocks/class-counter-controller.php
```diff
@@ -0,0 +1,104 @@
...
+            shortcode_ui_register_for_shortcode('shortcake_' . self::BLOCK_NAME, $shortcode_ui_args);
+        }
+
+        public function prepare_data($fields, $content, $shortcode_tag): array {
+            return [
+                'fields' => [
+                    'style' => $fields['style'],
+                    'completed' => $fields['completed'],
+                    'target' => $fields['target'],
+                    'percent' => round($fields['completed'] / $fields['target'] * 100),
```
kirdia: I think it is good to add some checks here to determine if values are set to prevent some php warnings/errors.
Or you can use [shortcode_atts](https://developer.wordpress.org/reference/functions/shortcode_atts/) to set default values if any field is not set.

tamlyn:
I have updated the PR to incorporate feedback and fix code style issues. Following discussion with @LauraHilliger this morning I have also had a stab at applying the visual design from the campaigns but it will need revisiting at some stage.

counters-mobile

Still outstanding:
- the thumbnail for the block
- the icons for the three counter styles

tests/unit/test-counter.php
```diff
@@ -0,0 +1,82 @@
+<?php
+/**
+ * PHP unit test of content four column block
```
kirdia: Please change the comment here for the right block

tests/unit/test-counter.php
```diff
@@ -0,0 +1,82 @@
...
+
+use P4BKS\Controllers\Blocks\Counter_Controller;
+use P4BKS\Views\View as View;
+
+require_once __DIR__ . '/../p4-unittestcase.php';
+
+if ( ! class_exists( 'P4_CounterTest' ) ) {
+
+	/**
+	 * Class P4_ContentFourColumnTest
```
kirdia: same here for the class name

angtheod:
Yes, let's wait for the missing images better.

classes/controller/blocks/class-counter-controller.php
```diff
@@ -0,0 +1,119 @@
...
+		/**
+		 * Get all the data that will be needed to render the block correctly.
+		 *
+		 * @param array  $fields This is the array of fields of the block.
+		 * @param string $content This is the post content.
+		 * @param string $shortcode_tag The shortcode tag of the block.
+		 *
+		 * @return array The data to be passed in the View.
+		 */
+		public function prepare_data( $fields, $content, $shortcode_tag ): array {
```
angtheod: a minor wpcs rule for missing space after closing parenthesis

PR #424, opened 2018-12-13
PLANET-3008 Removed blocks: Take Action, Campaign Thumbnails and CFC blocks

sagarsdeshmukh:
The Take Action, Campaign Thumbnails and CFC blocks are replaced by new covers block, hence we hide them on page 'Add page element' model pop up.

**Note:** On tag pages these blocks are still in use.


angtheod:
Shouldn't we remove also the code/templates for those blocks?
What about the places where those blocks are code-generated? @koyan ?

kirdia:
We want to keep these shortcodes for the time being, we just want to hide them from the ui, so that editors can't use them.
But I think it will break 'preview' mode in edit post, can you check @sagarsdeshmukh ?

sagarsdeshmukh:
> We want to keep these shortcodes for the time being, we just want to hide them from the ui, so that editors can't use them.
> But I think it will break 'preview' mode in edit post, can you check @sagarsdeshmukh ?

I checked for the page, whom these blocks were previously added, In this case, the preview displays shortcake tag as it is (without preview), but it works fine on the frontend. (Am I follow your point correctly @kirdia  ?)

koyan:
> But I think it will break 'preview' mode in edit post
We will not try to do anything about the preview in edit post. Editors should have removed/replaced these blocks in pages long before now. We keep the code only for the automatic tag pages.

PR #430, opened 2018-12-14
PLANET-3035 - Added alt_text to happy point block

oekeur:
For accessibility reasons all images should have an alt_text attribute set.
This commit makes sure the happy point tries to fetch an alt_text from the media library and otherwise adds a default generic alt_text

### DIFF
--- a/classes/controller/blocks/class-happypoint-controller.php
+++ b/classes/controller/blocks/class-happypoint-controller.php
@@ -154,6 +154,7 @@ if ( ! class_exists( 'HappyPoint_Controller' ) ) {
 			$fields['engaging_network_id'] = $options['engaging_network_form_id'] ?? '';
 			$fields['opacity']             = $opacity;
 			$fields['default_image']       = get_bloginfo( 'template_directory' ) . '/images/happy-point-block-bg.jpg';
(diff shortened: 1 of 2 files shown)


classes/controller/blocks/class-happypoint-controller.php
```diff
@@ -154,6 +154,7 @@ public function prepare_data( $fields, $content = '', $shortcode_tag = 'shortcak
 			$fields['engaging_network_id'] = $options['engaging_network_form_id'] ?? '';
 			$fields['opacity']             = $opacity;
 			$fields['default_image']       = get_bloginfo( 'template_directory' ) . '/images/happy-point-block-bg.jpg';
+			$fields['background_alt']      = get_post_meta( $fields['background'], '_wp_attachment_image_alt', true ) ?? 'Happy point background image';
```
kirdia: Hello @oekeur. Better use $image_id in this case, as it would take into account the default happypoint image. Also get_post_meta (with single true) can return an empty string, so it will apply the empty string when you use coalesce operator.
`
$image_alt = get_post_meta( $image_id, '_wp_attachment_image_alt', true );`
`$fields['background_alt'] = empty( $image_alt)  ? 'Happy point background image' : $image_alt ;`

kirdia:
Moved this to #439

PR #432, opened 2018-12-14
PLANET-1491: columns block

tamlyn:
This creates a new **Columns** block which combines the functionality of the 2 column, Static 4 Column and Take Action Tasks blocks into a single block. @mattkendon started it and I took it over so he could focus on Selenium work.

## Admin

The admin UI combines the fields necessary for all blocks, repeated 4 times to support up to 4 columns. Not all blocks originally supported all features (e.g. 2 column block had no images and only the tasks block had a block description). **Question**: Is it better to hide the irrelevant inputs in the UI depending on the selected block type, or update the templates and CSS to allow all content to be rendered in any block type?

## Frontend

I have combined the twig templates and styles of the two and four columns blocks into the the new block since they are quite similar. This should make things easier to maintain. The take action block is more complicated (with a separate layout and JS behaviour for mobile) so I didn't merge it in. Instead I `include` the (slightly modified) tasks template from the columns template and keep the original styles.

### Regressions 

Combining the two and four column styles has caused a few style regressions.

image

- At one breakpoint (between medium and large, see above) the original two column layout goes to one column but the new column block stays at two columns (like the four column block). Options: 
  - Use conditional in the template to switch the bootstrap classes used depending on block type
  - Don't use bootstrap override bootstrap styles at this breakpoint
- The original two column template has the column content arranged in rows. This would make it quite hard to follow in a screenreader which would read out all the titles, followed by all the column bodies, followed by all the buttons. However keeping the content horizontally aligned across columns is hard to do now (e.g. see buttons above). Options:
  - Use CSS grid but some form of fallback would be required for IE

In each case there is also the option of accepting the new behaviour, or reverting the template merge and using the original styles and templates instead (and only merging the admin portion of the block).

## Backward compatibility of data

This is a new block type but it modifies the existing `tasks` template to use the updates attribute names which will break existing usages of the Tasks block. **Question**: ~~Should I duplicate the tasks template to avoid this breakage? Or concentrate on the block migration `wp cli` work which is also required for the gallery block?~~ As suggested I have duplicated the `tasks.twig` template to avoid this breakage.


kirdia:
> ## Backward compatibility of data
> This is a new block type but it modifies the existing `tasks` template to use the updates attribute names which will break existing usages of the Tasks block. **Question**: Should I duplicate the tasks template to avoid this breakage? Or concentrate on the block migration `wp cli` work which is also required for the gallery block?

Probably the old tasks block and this new block will need to function for some time together, so better use a separate twig template for the tasks part of this block.

classes/controller/blocks/class-columns-controller.php
```diff
@@ -0,0 +1,217 @@
...
+					"description_$i" => $attributes[ "description_$i" ] ?? '',
+					"attachment_$i"  => $attributes[ "attachment_$i" ] ?? '',
+					"cta_text_$i"    => $attributes[ "cta_text_$i" ] ?? '',
+					"link_$i"        => $attributes[ "link_$i" ] ?? '',
+				];
+				$attributes_temp = array_merge( $attributes_temp, $column_atts );
+			}
+			$attributes = shortcode_atts( $attributes_temp, $attributes, $shortcode_tag );
+
+			$image_size = static::LAYOUT_TASKS === $attributes['columns_block_style'] ? 'medium' : 'thumbnail';
```
kirdia: Maybe it has meaning to use different image size depending on how many columns are defined here.
For example if you use 'tasks' type and specify one or 2 columns the images are distorted. 
But it's up to design to answer this (or the images could be constrained so they won't use the container's full width)

kirdia:
>  **Question**: Is it better to hide the irrelevant inputs in the UI depending on the selected block type, or update the templates and CSS to allow all content to be rendered in any block type?
I'm not sure about this. Probably general title and general description are needed for every block type, but it's up to the requestor for this to answer.

classes/controller/blocks/class-columns-controller.php
```diff
@@ -0,0 +1,217 @@
...
+					'type'  => 'textarea',
+					'meta'  => [
+						'placeholder' => __( 'Enter columns description', 'planet4-blocks-backend' ),
+						'data-plugin' => 'planet4-blocks',
+					],
+				],
+			];
+
+			// This block will have 4 different columns with same fields.
+			for ( $i = 1; $i < 5; $i++ ) {
```
angtheod: Better to add a MAX_COLUMNS constant for this.

classes/controller/blocks/class-columns-controller.php
angtheod: Indeed if we use lets say only 2 Tasks then these 2 images will get distorted and try to use all available width. After talking with William this is rather a css issue.
@comzeradd Do you think you can spot it?

kirdia:
- Rebased on release/1.29.
- Added js functionality for adding/removing columns.
- Added logic to use different image size based on layout and number of columns.

admin/js/blocks-ui.js
```diff
@@ -489,6 +489,133 @@ var p4_blocks = {
...
+          columns.show_row(++row);
+          $element.parent().data('row', row);
+          $('.shortcake-columns-remove-column').removeAttr('disabled');
+          if (row === 4) {
+            $element.attr('disabled', 'disabled');
+          }
+        }
+      });
+
+      $('.shortcake-columns-remove-column').on('click', function (event) {
```
angtheod: @kirdia Nice addition and it works visually but when clicking Update button there is no ajax request `bulk_do_shortcode` so the block's shortcode does not actually update. Is that right or am I doing something wrong?

kirdia: I can't reproduce it! When I'm adding/removing a row it does an ajax request ('bulk_do_shortcode'). You have to change at least one field though to trigger it.

angtheod: How can we trigger it in case we just edit block and delete a column (without changing other fields)?

kirdia: Deleting a row will clear the column's fields values, so it should trigger the ajax request.
But if the row's fields were already empty it won't.

I'll look it again, I think the shortcake's models should be updated also to reflect the changes in the shortcode and not just clear the html fields.

PR #439, opened 2018-12-20
PLANET-3035 Added alt_text to happy point block

kirdia:
Moving from #430 PR
For accessibility reasons all images should have an alt_text attribute set.
This commit makes sure the happy point tries to fetch an alt_text from the media library and otherwise adds a default generic alt_text


classes/controller/blocks/class-happypoint-controller.php
```diff
@@ -148,12 +148,14 @@ public function prepare_data( $fields, $content = '', $shortcode_tag = 'shortcak
...
 			$image_id                      = '' !== $fields['background'] ? $fields['background'] : $p4_happy_point_bg_image;
 			$img_meta                      = wp_get_attachment_metadata( $image_id );
+			$image_alt                     = get_post_meta( $image_id, '_wp_attachment_image_alt', true );
 			$fields['background_src']      = wp_get_attachment_image_src( $image_id, 'retina-large' );
 			$fields['background_srcset']   = wp_get_attachment_image_srcset( $image_id, 'retina-large', $img_meta );
 			$fields['background_sizes']    = wp_calculate_image_sizes( 'retina-large', null, null, $image_id );
 			$fields['engaging_network_id'] = $options['engaging_network_form_id'] ?? '';
 			$fields['opacity']             = $opacity;
 			$fields['default_image']       = get_bloginfo( 'template_directory' ) . '/images/happy-point-block-bg.jpg';
+			$fields['background_alt']      = empty( $image_alt ) ? 'Happy point background image' : $image_alt;
```
angtheod: Perhaps an alt text without block name and also translatable?

kirdia: Updated!

kirdia:
Updated pot template also to include new string.

PR #440, opened 2018-12-20
Submenu update

tamlyn:
https://jira.greenpeace.org/browse/PLANET-1766
https://jira.greenpeace.org/browse/PLANET-3072

## Changes
- Support three heading levels instead of two
- Support configurable list style (none, bullet, number)
- Change to single `ul` for whole menu instead of one `ul` per top level item. This should improve accessibility and also makes the numbering of items easier.
- User configurable nesting level. On the mockups this is done through a separate dropdown however after discussion with @LauraHilliger last week, removed the explicit dropdown in favour of implicitly increasing the nesting level for links only. 

### Updated menu with links and bullets

menu-link-updated

### Updated menu no bullets or links on top level

menu-nolink-updated

### Updated menu with three nesting levels and numbers

menu-three-levels

  
## Layout

Note that I haven't changed the flexbox based row layout. It works in the above case because the page has two main sections of comparable length. However on the the [NRO index page](https://www.greenpeace.org/international/worldwide/) it results in the items being arranged left to right then top to bottom (rather than top to bottom then left to right as I would expect). On pages with a less uniform heading distribution it's even worse. Here's the original submenu block on the default content for the privacy page:

menu-link-original

One possible solution is to arrange the content using CSS columns. This matches the wireframes on the original ticket but can lead to situations where one menu item is split across multiple columns. For example:

menu-columns

I feel this is still an improvement over the current layout for most cases. Those changes are available in [a branch](https://github.com/greenpeace/planet4-plugin-blocks/tree/submenu-update) that I worked on before checking whether this was within scope for the ticket.


classes/controller/blocks/class-submenu-controller.php
```diff
@@ -198,58 +196,89 @@ public function prepare_data( $attributes, $content = '', $shortcode_tag = 'shor
...
-						// Break if we get to next heading.
-						// phpcs:ignore WordPress.NamingConventions.ValidVariableName
-						if ( $node->nodeName === $heading_tag ) {
-							break;
-						}
+		/**
+		 * Process flat array of DOM nodes to build up menu tree structure.
+		 *
+		 * @param int   $current_level Current menu nesting level.
+		 * @param array $nodes Array of heading DOM nodes, passed by reference.
```
angtheod: Could we define $nodes as `\DOMNode[]` so that our editors can find the fields that we are trying to access later on.

angtheod:
Not a part of this PR but I am getting js error
`Uncaught TypeError: Cannot read property 'top' of undefined` for this [line](https://github.com/greenpeace/planet4-plugin-blocks/blob/submenu-update/admin/js/submenu.js#L54)

kirdia:
> Not a part of this PR but when clicking on one of the submenu links I am getting js error
> `Uncaught TypeError: Cannot read property 'top' of undefined` for this [line](https://github.com/greenpeace/planet4-plugin-blocks/blob/submenu-update/admin/js/submenu.js#L54)

It is caused from this change
https://github.com/greenpeace/planet4-plugin-blocks/pull/440/files#diff-db0c911c0bf0c34109437131635e9458R301

Actually no anchor is created in the frontend. Do the menu links work for you?

angtheod:
> Actually no anchor is created in the frontend. Do the menu links work for you?

The heading type was not passed to js so the anchor links were not being created.
Also, resolved conflicts.

angtheod:
Fixed flickering issue when clicking on submenu link due to momentarily navigating to hash id right before starting the scrolling. thanks @kirdia

PR #445, opened 2019-01-10
Planet-2561 - Added customizations for campaign post types

mehulkaklotar:
Ref: https://jira.greenpeace.org/browse/PLANET-2561


onyekaa:
I've merged develop into this branch due to the changes made that cause conflicts with the work done so far. All theme styles and images from outlandish have been added in this PR.

assets/scss/style.scss
```diff
@@ -2,6 +2,7 @@
 @import "base/variables";
 @import "base/functions";
 @import "base/mixins";
+@import "base/blocks";
```
comzeradd: This is a rather huge file. This should probably needs to break down into smaller components.

assets/scss/base/_mixins.scss
```diff
@@ -187,3 +187,97 @@ $nested-comment-left-margin: $n50;
     }
   }
 }
+
+@mixin monochrome {
```
comzeradd: Let's avoid adding a new set of media queries mixins. You can use the ones already in place and adjust your code to be mobile first.

assets/scss/themes/_theme_antarctic.scss
```diff
@@ -0,0 +1,259 @@
+@import "../base/mixins";
+@import "../base/variables";
+@import "../base/blocks";
```
comzeradd: Since we will include all themes on `style.scss`, why do you re-import these here?

assets/scss/themes/_theme_antarctic.scss
```diff
@@ -0,0 +1,259 @@
...
+
+@font-face {
+  font-family: sanctuary;
+  src: url("fonts/antarctic/sanctuary.otf");
+}
+
+body.theme-antarctic {
+  @import url("https://fonts.googleapis.com/css?family=Montserrat:400,500,600,800");
```
comzeradd: Please create a `_fonts.scss` file under `base/` for all font declarations. See [this example](https://github.com/greenpeace/planet4-master-theme/blob/develop/assets/scss/base/_fonts.scss).

assets/scss/themes/_theme_antarctic.scss
```diff
@@ -0,0 +1,259 @@
+@import "../base/mixins";
+@import "../base/variables";
+@import "../base/blocks";
+
+@font-face {
+  font-family: sanctuary;
+  src: url("fonts/antarctic/sanctuary.otf");
```
comzeradd: I noticed you placed this font under `images`. It's probably better to go under `assets/fonts`.

assets/scss/style.scss
onyekaa: Hi. How do you suggest I split them? By block types?

comzeradd: Yes, but as far as I can see this includes blocks that already exist. So the code should be split into the right components.

You use mixins to redefine some existing blocks, creating a lot of code repetition. It's not clear to me why this approach was chosen to suggest an alternative. But this will make the blocks prone to regression bugs if we need to do any change in the future. Maybe define the mixins right in their existing component, with some default values that you can override when you include them for the campaigns.

onyekaa: I didn't write this, this is the code gotten from Outlandish, so maybe we can discuss this with them?

mcshanea:
@comzeradd squashing commits is usually quite a painful exercise, it moves us backwards in testing, and we can't invest continuous development time. Could you either jump in and do this yourself or we commit as is.

comzeradd:
@mcshanea this is a standard review process for opening and merging PRs. Squashing would have been avoided if the feature branch was initially rebased, instead of merging. We can't really merge a PR with merge commits from the target branch.

See the git best practices that we tend to follow on planet4 (especially the "Amend" and "Rebase" parts).
https://planet4.greenpeace.org/handbook/dev-git-best-practices/

onyekaa:
@comzeradd Commits have been squashed.

assets/scss/style.scss
tomisaacmarshall: Hey guys, re the approach taken for the campaign styles ... my understanding is that we were provided with flat html and css for a showcase mock-up page and were asked to write css to apply the new designs to this html without altering the html or base css. This was before we were involved with working on the planet4 blocks plugin itself. Hope that's helpful?

comzeradd:
Closing this in favor of #526

PR #446, opened 2019-01-10
Upgrade frontend libraries

comzeradd:
Similar to greenpeace/planet4-master-theme#684


comzeradd:
Interesting. Maybe there is a different behavior between different npm versions.

PR #448, opened 2019-01-17
Update circleci config

kirdia:
Use repo's dev dependencies versions (phpcs, phpunit), instead of configuring them in ci.


comzeradd:
@kirdia can you also add `node` upstream repo? This seems to cause the entire pipeline fail at the moment. Similar to [what we have](https://github.com/greenpeace/planet4-master-theme/blob/develop/.circleci/config.yml#L28-L29) on master-theme.

PR #455, opened 2019-02-05
PLANET-3179: Implement new Carousel Header design

pablocubico:
Ref: https://jira.greenpeace.org/browse/PLANET-3179

I don't have a design for: 
`a) Create a style chooser on the existing carousel Header block`

So I did this:
image

I avoided any changes that could break the existing carousel or doing massive opinionated refactors to the blocks flow. For the future, we could think of some refactor, for example, to allow different templates, styles & javascripts for a single block.


classes/controller/blocks/class-carouselheader-controller.php
```diff
@@ -21,6 +21,9 @@ class CarouselHeader_Controller extends Controller {
 		/** @const string BLOCK_NAME */
 		const BLOCK_NAME = 'carousel_header';
 
+		const LAYOUT_ZOOM_AND_SLIDE_TO_GRAY = 'zoom-and-slide-to-gray';
+		const LAYOUT_FULL_WIDTH_CLASSIC 	= 'full-width-classic';
```
kirdia: Alignment off here..

classes/controller/blocks/class-carouselheader-controller.php
```diff
@@ -33,7 +36,28 @@ class CarouselHeader_Controller extends Controller {
...
+							'image' => esc_url( plugins_url() . '/planet4-plugin-blocks/admin/images/carousel-with-preview.png' ),
+						],
+						[
+							'value' => static::LAYOUT_FULL_WIDTH_CLASSIC,
+							'label' => __( 'Full width classic', 'planet4-blocks-backend' ),
+							'desc'  => 'This is a full width slider with a classic look: big slides, fade transition.',
+							'image' => esc_url( plugins_url() . '/planet4-plugin-blocks/admin/images/carousel-classic.png' ),
+						],
+					],
+				]
```
kirdia: There are 2 code inspection errors in ci jobs

comzeradd:
On wp-admin, it seems that the images indicating the two different options are big enough to overlap.

carousel1

comzeradd:
You could leave a bit more padding for the captions section. Currently it overlaps with the arrow icons on some resolutions.

carousel2

comzeradd:
When changing slides using the indicators at the bottom there is no animation. Is this by design?

There is also this weird effect, where the cta button from the other slide gets visible. See screencast:

carousel

pablocubico:
@comzeradd 
Admin side: weird, I explicitly added styles to fix that, wlll check!
Arrow overlap: we currently have this arrow overlap problem with RTL sites where the headers & buttons are on the right side, I'll see what I can do about it. 
Transitions: Well, there was no comments in the spec regarding this, so I avoided adding more code to achieve that in order to reuse the most of the current one, which disables transitions except for "adjacent" slides. 
Weird effect: wow, that's an awful bug, will check.

When I started, I followed the indication of the ticket, which is to create a styling option on the current Carousel Header block, now I think it should have been a "block" of its own, given the many differences in style & behavior, but that would prevent the users from magically switching from one style to the other in any given block, this approach may carry a few trade-offs. That said, I think that with a little tweaking, and fixing the things you mentioned, we'll get the desired result.

koyan:
> When I started, I followed the indication of the ticket, which is to 
> create a styling option on the current Carousel Header block, now 
> I think it should have been a "block" of its own
No, in theory if the one affects the other, you should just create a second twig file and call that one as well, instead of trying to make both work with the same twig file

pablocubico:
@koyan  Well, it's not just Twig, it's twig + JS + CSS. The way things are built now are pretty tied up to the `BLOCK_NAME` constant: https://github.com/greenpeace/planet4-plugin-blocks/blob/develop/classes/controller/blocks/class-controller.php#L158

I could use an `include` statement in the main Twig file (`carousel_header`), in order to split the code, but then again, we also have the styles and the JS logic. 

I thought the goal of this was to make it a "drop-in" replacement for the current one, which works quite well as it is (despite the bugs found by Nikos). If we want to make it cleaner code-wise, I guess we'll have to refactor `class-controller` to allow overriding the `BLOCK_NAME` value at runtime with another value in order to choose which template + styles + code to render depending on the chosen style. The other option would be of course making it a totally new block.  But as I said, I didn't want to engage in refactors that could affect the way all blocks work. So I think the current version works, as long as I fix the bugs. 

Let me know what you think. Pinging @comzeradd and @kirdia too for ideas.

pablocubico:
> On wp-admin, it seems that the images indicating the two different options are big enough to overlap.
> 
> carousel1

@comzeradd I think this may be a caching issue, can't reproduce it locally. Can you try flushing th cache? Thanks!

pablocubico:
> When changing slides using the indicators at the bottom there is no animation. Is this by design?
> 
> There is also this weird effect, where the cta button from the other slide gets visible. See screencast:
> 
> carousel

Ok, so this overflowing bug is happening with the current carousel too, if you add too much text on the `Subheader` and `description` fields. 

Things to consider:
As the carousel needs to have a fixed height, we should set a max length for the subheader (in fact, for all the fields...), since its shown as a text (not a textarea) in the admin side:

image

Also, I just noticed there are only two texts in the design, no subheader:

image

I don't know what to do about that, should we hide the subheader? What if it's in use in an NRO?

pablocubico:
> You could leave a bit more padding for the captions section. Currently it overlaps with the arrow icons on some resolutions.
> 
> carousel2

I narrowed the containers for the arrows and added a little padding in one of the responsive breakpoints. It's quite impossible to cover all the cases, I think we should consider only a set of realistic screensizes, that actual devices use, and not all the intermediate sizes, or we'll need to use a custom set of breakpoints for this particular scenario, at least with the current design.

Let me know how it looks now.

pablocubico:
Updates:

- Fixed code styling issue.
- Arrows and arrow areas are narrower and work in most resolutions.
- Responsive polishing.
- Rebased with latest develop.

kirdia:
> @koyan Well, it's not just Twig, it's twig + JS + CSS. The way things are built now are pretty tied up to the `BLOCK_NAME` constant: https://github.com/greenpeace/planet4-plugin-blocks/blob/develop/classes/controller/blocks/class-controller.php#L158
> 
> I could use an `include` statement in the main Twig file (`carousel_header`), in order to split the code, but then again, we also have the styles and the JS logic.
> 
> I thought the goal of this was to make it a "drop-in" replacement for the current one, which works quite well as it is (despite the bugs found by Nikos). If we want to make it cleaner code-wise, I guess we'll have to refactor `class-controller` to allow overriding the `BLOCK_NAME` value at runtime with another value in order to choose which template + styles + code to render depending on the chosen style. The other option would be of course making it a totally new block. But as I said, I didn't want to engage in refactors that could affect the way all blocks work. So I think the current version works, as long as I fix the bugs.
> 
> Let me know what you think. Pinging @comzeradd and @kirdia too for ideas.


I don't see why we a change in the block name is essential here.
Could you elaborate more this, in case I'm missing something?

As long as the block's input data remain same, I don't see a need for that change.

If there are too many design/functionality changes needed, then we can go with a different pair of twig/markup, css. 
I guess it would result in some duplication in css, but I don't know how we can avoid that.
If there are not too many changes needed in markup, then you can add some logic in the existing twig and change css, so they won't interfere each other.

pablocubico:
> > @koyan Well, it's not just Twig, it's twig + JS + CSS. The way things are built now are pretty tied up to the `BLOCK_NAME` constant: https://github.com/greenpeace/planet4-plugin-blocks/blob/develop/classes/controller/blocks/class-controller.php#L158
> > I could use an `include` statement in the main Twig file (`carousel_header`), in order to split the code, but then again, we also have the styles and the JS logic.
> > I thought the goal of this was to make it a "drop-in" replacement for the current one, which works quite well as it is (despite the bugs found by Nikos). If we want to make it cleaner code-wise, I guess we'll have to refactor `class-controller` to allow overriding the `BLOCK_NAME` value at runtime with another value in order to choose which template + styles + code to render depending on the chosen style. The other option would be of course making it a totally new block. But as I said, I didn't want to engage in refactors that could affect the way all blocks work. So I think the current version works, as long as I fix the bugs.
> > Let me know what you think. Pinging @comzeradd and @kirdia too for ideas.
> 
> I don't see why we a change in the block name is essential here.
> Could you elaborate more this, in case I'm missing something?
> 
> As long as the block's input data remain same, I don't see a need for that change.
> 
> If there are too many design/functionality changes needed, then we can go with a different pair of twig/markup, css.
> I guess it would result in some duplication in css, but I don't know how we can avoid that.
> If there are not too many changes needed in markup, then you can add some logic in the existing twig and change css, so they won't interfere each other.

No, no, it's not essential, on the contrary, I don't want to mess with the block name! And as you said, the data input is the same, I wanted this to be a drop-in replacement for the current carousel (I think that was the goal of the issue). That's why I reused the current code with some differences in the JS file. That said, I think it would be cleaner to have separate files for different layouts if the carousels are too different, instead of having every layout style in the same twig file and logic differences in the same JS file, or in case we want to add more logic in the future.

But as for now, I think it's ok the way it is, I did a few tweaks yesterday for better responsiveness and it's looking good to me.

comzeradd:
I think it's good to go, and we could iterate later based on feedback.

Please squash the commits. At least 4906b35 should be squashed.

koyan:
Well, the phpcs tests still fail as far as I Can see, and the styles need to be merged again

comzeradd:
:100:

PR #456, opened 2019-02-06
PLANET-3158: Explore page spacing off after split two col block

pablocubico:
Added a little margin on the top of the articles block.

Ref: https://jira.greenpeace.org/browse/PLANET-3158


comzeradd:
@pablocubico switch that to the release branch?

PR #465, opened 2019-02-14
PLANET-3223: New Carousel block corrections

pablocubico:
Note: this shows the description only in the new Carousel Header, not in the old one.

Ref: https://jira.greenpeace.org/browse/PLANET-3223

This also fixes: https://jira.greenpeace.org/browse/PLANET-3220

(I got confused with the ticket numbers, but it's all related).


comzeradd:
Code looks good. Nice work :+1: 

Since the CTA button has fixed width, we can limit the text from overflowing.

button

We do something similar to other Blocks buttons. Something like this should work:

```
white-space: nowrap;
overflow: hidden;
text-overflow: ellipsis;
```

comzeradd:
Did we also concluded with the text limiting discussion? The CTA button overlaps with the indicators when the text is too long.

cta

This Carousel style is longer than the old one. So, maybe we can lift or increase the width restriction we have on header and subheader. Unless this is a design requirement.

pablocubico:
@comzeradd there is a limit now in the admin page: 32 chars for title, 200 for the description.

I'll try the ellipsis thing!

PR #466, opened 2019-02-15
PLANET-3206 Fix empty string supplied PHP warning in Submenu block

sagarsdeshmukh:


classes/controller/blocks/class-submenu-controller.php
```diff
@@ -217,7 +217,9 @@ private function parse_post_content( $content, $attributes ) {
 			}
 
 			$dom = new \DOMDocument();
-			$dom->loadHtml( $content );
+			if ( '' === $content ) {
```
angtheod: If I am not mistaken we should make sure that $content is not empty.
We could do something like
`if ( ! $content ) { return []; }`

angtheod:
I get a couple more warnings when I run `wp elasticpress index`.

One is in submenu block again [here](https://github.com/greenpeace/planet4-plugin-blocks/blob/release/v1.33.0/classes/controller/blocks/class-submenu-controller.php#L184) and we need to check if $post is an object like this:
`$content = $post instanceof \WP_Post ? $post->post_content : $content;`

the other one is in engagingnetworks plugin [here](https://github.com/greenpeace/planet4-plugin-engagingnetworks/blob/develop/classes/controller/blocks/class-enform-controller.php#L350) and we need to check that array has at least 2 elements like this:
`if ( $attr_parts_num >= 2 && is_numeric( $attr_parts[0] ) && 'mandatory' !== $attr_parts[ $attr_parts_num - 1 ] ) {`

classes/controller/blocks/class-submenu-controller.php
kirdia: Yes, I agree with @angelos on this. No need to continue execution if no content is set.
Returning an empty array at this point would be enough.

angelos: Thanks for including me in the conversation! However, I'm pretty sure you meant to mention @angtheod :)

kirdia: hahaha :smile:  yes, true. it was meant for @angtheod. sorry for the false notification :smiley:

angtheod:
Please fix also the PHPCS error.

PR #467, opened 2019-02-15
PLANET-3034 - Reorder plugin blocks report to be alphabetical

koyan:
Relevant Jira issue [PLANET-3034](https://jira.greenpeace.org/browse/PLANET-3034)

Reorder plugin blocks report to be alphabetical
And fix a couple of phpcs errors

### DIFF
--- a/planet4-blocks.php
+++ b/planet4-blocks.php
@@ -165,12 +165,14 @@ function plugin_blocks_report() {
 	// phpcs:disable
 	foreach ( $blocks as $block ) {
 		$block = substr( $block, 10 );
-		$sql = 'SELECT ID, post_title
-				FROM `wp_posts` 
-				WHERE post_status = \'publish\' 
-				AND `post_content` LIKE \'%[shortcake_' . $block . '%\'
-				ORDER BY post_title
-				';
+		$shortcode = '%[shortcake_' . $wpdb->esc_like( $block ) . '%';
+		$sql       = $wpdb->prepare(
+			"SELECT ID, post_title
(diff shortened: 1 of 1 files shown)


angtheod:
Nice, I think [here](https://github.com/greenpeace/planet4-plugin-blocks/pull/467/files#diff-251b93d9ce779a235999494d335a1648R183) we could also remove the underscores (as requested in description) and capitalize first letter by outputting this `ucfirst( str_replace( '_', ' ', $block ) )`

planet4-blocks.php
```diff
@@ -160,25 +160,27 @@ function plugin_blocks_report() {
 
 	// Array filtering on shortcake shortcodes.
 	$blocks = array_filter( array_keys( $shortcode_tags ), 'is_shortcake' );
+	sort( $blocks );
 
 	// phpcs:disable
 	foreach ( $blocks as $block ) {
-		$block = substr($block, 10);
+		$block = substr( $block, 10 );
```
angtheod: Just to be on the safer side I think we should sql escape what is in $block.
We could use esc_like() and use result in a prepared statement like this
```$block     = substr( $block, 10 );
		$shortcode = '%[shortcake_' . $wpdb->esc_like( $block ) . '%';
		$sql       = $wpdb->prepare(
			"SELECT ID, post_title
			FROM `wp_posts` 
			WHERE post_status = 'publish' 
			AND `post_content` LIKE %s
			ORDER BY post_title",
			$shortcode );```

koyan: Done.

PR #468, opened 2019-02-15
PLANET-3227 remove old blocks

koyan:
[Related Jira Issue](https://jira.greenpeace.org/browse/PLANET-3227)

Making the 5 blocks that we have replaced not available in the interface, while leaving them working (for any pages that still have them).

This is temporary, until the NROs replace their existing blocks with the new ones, in which time we will completely remove these 5 from our codebase


classes/controller/blocks/class-carousel-controller.php
```diff
@@ -100,7 +100,7 @@ public function prepare_fields() {
 				'label'         => __( 'Carousel', 'planet4-blocks-backend' ),
 				'listItemImage' => '<img src="' . esc_url( plugins_url() . '/planet4-plugin-blocks/admin/images/take_action_carousel.png' ) . '" />',
 				'attrs'         => $fields,
-				'post_type'     => P4BKS_ALLOWED_PAGETYPE,
+				'post_type'     => 'do_not_show_replaced_by_gallery',
```
sagarsdeshmukh: The `page_type` parameter would be expected as array, can you please make it - `['do_not_show_replaced_by_gallery']`

PR #474, opened 2019-02-19
PLANET-3194: Columns Block: issues with No Image and Icons styles

pablocubico:
Addressing fixes from UAT:

Note: didn't fix point 2 on the "No image" part as I don't know what the spec is for the font size.

No Image
- [x] 1) Mobile: button alignment is off
- [ ] 2) Header is wrong font size; unless this change was by design, William Morris-Julien?
- [x] 3) Editor should not be able to generate a button when no (valid) URL is entered

Icon Style
- [x] 1) Hyperlinked text in the Description field is the wrong font - should be Lora, not Roboto.
- [x] 2) if Editor generates a CtA when no (valid) URL is entered, it should not be hyperlinked

Tasks Style
- [x] 1) Mobile: button alignment is off
- [x] 2) Image is hyperlinked even when no URL appears in the button.


pablocubico:
@comzeradd I don't know how to amend a particular commit, but I rebased the two into one and added a more descriptive message. 👍

PR #476, opened 2019-02-20
PLANET-2338 - new Carousel Header: Fix background focus

comzeradd:


pablocubico:
@comzeradd Approved sir! Just needs new minified files.

PR #500, opened 2019-03-07
Set blocks to work on campaign post type.

onyekaa:
Since the styling situation is pending, I created a new branch to just initialize the blocks on campaign post types. Once the other stuff is settled I'll merge, but for testing on the staging site, it's imperative that we can add least add blocks.


kirdia:
@onyekaa please use a specific branch as your base branch, so the work for campaigns won't interfere with weekly releases we do, the same way we did for master-theme
https://github.com/greenpeace/planet4-master-theme/pull/708

should I create a planet-2561 branch for blocks repo also?

PR #507, opened 2019-03-12
PLANET 3271: Check character truncation caused by the button width difference between the two homepage carousels

pablocubico:
Ref: https://jira.greenpeace.org/browse/PLANET-3271


kirdia:
Looks good! @pablocubico please fix the style conflicts

PR #513, opened 2019-03-19
PLANET-2978: Articles block: article count allowing for minus numbers

dpivo:
The code has been changed to default the number of articles to 3 if the user inputs less than 0 in the number of articles field.

The corresponding unit test was also updated to test for this scenario.

Ref: https://jira.greenpeace.org/browse/PLANET-2978


kirdia:
@dpivo there are some code inspection errors there, please check them

you can try running the phpcs locally
```
 composer install
 vendor/bin/phpcs --config-set installed_paths vendor/wp-coding-standards/wpcs
 vendor/bin/phpcs 
```

Also would be a nice addition to add min, max to the input field directly.
https://github.com/greenpeace/planet4-plugin-blocks/blob/a38c896e35aea98a034af1d19fbc0eb1961d885a/classes/controller/blocks/class-articles-controller.php#L150
```
						'min' => 1,
						'max' => self::MAX_ARTICLES,
```

tests/unit/test-articles.php
```diff
@@ -134,6 +134,39 @@ public function test_story_count() {
...
+						$this->assertNotFalse( $res );
+					} catch ( \Exception $e ) {
+						$this->fail( sprintf( '->Unable to add term to post with id %d.', $id ) );
+					}
+				}
+			}
+
+			$fields = [
+				'article_count'      => -2,
+				'p4_page_type_story' => true,
```
kirdia: Nice additon the testcase!
This field is removed though (p4_page_type_story), it is from an older version.
It should be 'post_types'

You can check in the newcovers tests how you can create posts with a term of the custom taxonomy.
https://github.com/greenpeace/planet4-plugin-blocks/blob/develop/tests/unit/test-newcovers-controller.php#L251
https://github.com/greenpeace/planet4-plugin-blocks/blob/develop/tests/unit/test-newcovers-controller.php#L139

kirdia:
Good! but check also the inspection errors

PR #519, opened 2019-03-27
PLANET-3346: Gallery Block, Three columns style: Uses full images instead of thumbnails

pablocubico:
Ref: https://jira.greenpeace.org/browse/PLANET-3346

### DIFF
--- a/includes/blocks/gallery.twig
+++ b/includes/blocks/gallery.twig
@@ -96,7 +96,8 @@
                 <div class="col">
                   <div class="{{ ordinals[key] }}-column split-image">
                     {% if ( single_image.image_src ) %}
-                      <img srcset="{{ single_image.image_srcset }}"
+                      <img src="{{ single_image.image_src }}"
+                        srcset="{{ single_image.image_srcset }}"
                         sizes="{{ single_image.image_sizes }}"
                         alt="{{ single_image.alt_text }}">
                     {% endif %}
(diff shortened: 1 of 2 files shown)


includes/blocks/gallery.twig
```diff
@@ -96,7 +96,9 @@
                 <div class="col">
                   <div class="{{ ordinals[key] }}-column split-image">
                     {% if ( single_image.image_src ) %}
-                      <img src="{{ single_image.image_src }}" alt="{{ single_image.alt_text }}">
+                      <img srcset="{{ single_image.image_srcset }}"
+                        sizes="{{ single_image.image_sizes }}"
```
kirdia: I don't think multiple sources are needed for grid (and maybe for 3 columns).
The thing is that `wp_get_attachment_image_srcset` will provide source for all image sizes available.
I don't know if this will result in browser picking up a larger size than needed. Needs confirmation

pablocubico:
@kirdia it looks like it picks the correct image to me, for example:

In the develop branch:

Captura de pantalla 2019-03-29 a la(s) 10 12 13

In this branch:

Captura de pantalla 2019-03-29 a la(s) 10 13 13

kirdia:
Yes, sorry, ignore that. I just realized that each style has it's own markup.
Thank your for checking it out.

Just another one which is not part of this, but it's a small fix.
Can you add img src attribute for split-three-column also?
https://github.com/greenpeace/planet4-plugin-blocks/blob/5dc8601ca8d617a9254fd644f6d2b344d64cecd0/includes/blocks/gallery.twig#L99

pablocubico:
Updated!

PR #520, opened 2019-03-29
Fixes issue with timeline preview iframe not having css if the post h…

tomisaacmarshall:
Fixes issue with timeline preview iframe not having css if the post hasn't been saved

Also removes timeline max height as realised it wasn't really in keeping with other blocks


classes/controller/blocks/class-timeline-controller.php
```diff
@@ -102,15 +102,17 @@ class Timeline_Controller extends Controller {
...
 					global $post;
 
-					if ( strpos( $post->post_content, 'shortcake_' . self::BLOCK_NAME ) !== false ) {
-						$css = 'https://cdn.knightlab.com/libs/timeline3/' . self::TIMELINEJS_VERSION . '/css/timeline.css';
+					$has_timeline = strpos( $post->post_content, 'shortcake_' . self::BLOCK_NAME ) !== false;
 
+					// Load css only if timeline shortcode appears in post content, or the post hasn't been saved yet
+					// so we don't know (so as not to break the preview iframe before the post is saved).
+					if ( $has_timeline || 'auto-draft' === $post->post_status ) {
+						$css = 'https://cdn.knightlab.com/libs/timeline3/' . self::TIMELINEJS_VERSION . '/css/timeline.css';
```
kirdia: Thank you for the fix @tomisaacmarshall 

One last thing
Instead of checking for the post_status here
`|| 'auto-draft' === $post->post_status`


You can enqueue the style also in `prepare_template_preview_iframe` since you have overriden it already.
https://github.com/greenpeace/planet4-plugin-blocks/blob/release/v1.39.0/classes/controller/blocks/class-timeline-controller.php#L270

We'll be still some duplication with this solution (we will handle it from there), but it is better than en-queuing the style for all blocks.

tomisaacmarshall: Hmm, I just tried that and it didn't seem to work. That method is also called too late to enqueue it in the head, I think. Is it really an issue that this stylesheet would occasionally be enqueued by preview iframes that don't need it, only ever when the post is in auto-draft?

kirdia: Νο, it is not such a big issue

PR #521, opened 2019-03-29
PLANET-3372 Added report for different carousel header styles

koyan:
[Jira issue 3372](https://jira.greenpeace.org/browse/PLANET-3346)

This adds two rows to the plugins usage report about which style of the carousel Header block is used
In the admin pages of wordpress:
plugins-report-admin
And in the json for the automatic report
plugins-report

Additionally it fixes a bug that existed that was making the "Carousel" to also report all  "Carousel_header" blocks.

### DIFF
--- a/planet4-blocks.php
+++ b/planet4-blocks.php
@@ -167,7 +167,7 @@ function plugin_blocks_report() {
 	// phpcs:disable
 	foreach ( $blocks as $block ) {
 		$block = substr( $block, 10 );
-		$shortcode = '%[shortcake_' . $wpdb->esc_like( $block ) . '%';
+		$shortcode = '%[shortcake_' . $wpdb->esc_like( $block ) . ' %';
 		$sql       = $wpdb->prepare(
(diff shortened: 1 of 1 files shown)


pablocubico:
@koyan looks good to me, the only thing to comment is that the "zoom-and-slide-to-gray" block style is the default style and it may not be present in old shortcodes. Having a shortcode without `block_style` on it will still render the `zoom-and-slide-to-gray` style carousel header. This could be an issue if an NRO has an old shortcode somewhere and never updated their carousel headers. 

If you don't think this is a problem, I think it's good to go. 👍

koyan:
Oh, that is a very usefull point. The whole point of this report is to identify NROs with the old style! 
Thanks. I will write a different SQL to include shortcodes without block_style on Monday.

koyan:
Oh, that is a very usefull point. The whole point of this report is to identify NROs with the old style! 
Thanks. Report amended to include in the slide-to-gray anything that does not have a style defined.

PR #530, opened 2019-04-09
PLANET-3380 admin convert blocks

angtheod:
Add Menu Item for Blocks and a submenu item. The Blocks Usage logic was moved to the submenu item and in the main menu item we added a button and an input for Editors to use for replacing block shortcodes with new ones.

*`class-blocks-usage-controller.php` has the code that was in plugin-blocks.php (related to block usage report functionality)... so no real changes on this one, only restructured so that it becomes a submenu item.

### DIFF
--- a/classes/controller/menu/class-settings-controller.php
+++ b/classes/controller/menu/class-settings-controller.php
@@ -2,11 +2,13 @@
 /**
  * Settings class
  *
- * @package P4BKS
- * @since 0.1.0
+ * @package P4BKS\Controllers\Menu
+ * @since 1.40.0
  */
 
 namespace P4BKS\Controllers\Menu;
+
+use P4BKS\Command\ShortcodeReplacer;
 
 if ( ! class_exists( 'Settings_Controller' ) ) {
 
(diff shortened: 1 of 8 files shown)


angtheod:
Works with shell_exec, but I am trying to make it work with WP_CLI::runcommand for more safety.
image

classes/controller/menu/class-settings-controller.php
```diff
@@ -32,47 +34,84 @@ public function create_admin_menu() {
...
+
+			$data['nonce_action'] = 'convert-blocks-nonce';
+			$data['form_submit']  = 0;
+
+			if ( 'POST' === $_SERVER['REQUEST_METHOD'] ) {
+				$data['form_submit'] = 1;
+
+				if ( ! wp_verify_nonce( $nonce, 'convert-blocks-nonce' ) ) {
+					$data['message'] = __( 'Nonce verification failed!', 'planet4-blocks-backend' );
+					return false;
```
sagarsdeshmukh: If nonce verification fails, the error message will not display because of return false.

sagarsdeshmukh:
@angtheod - On my local env. the Rest API end point(https://www.planet4.test/wp-json/plugin_blocks/v1/plugin_blocks_report/) throughs 404. could you please confirm it once.

comzeradd:
Please make sure to squash the last two commits (the fixes). Either locally or here on merge.

PR #531, opened 2019-04-09
PLANET-3417 fix cover card scrolling

angtheod:
Fix Cover card not following scroll movement in some Posts.

I can tell this change covers more cases, but does it cover all?
@pablocubico @comzeradd


comzeradd:
It definitely expands the scope of this selector, so it should fix these corner cases. I can't think of a case that would escape this selector.

PR #533, opened 2019-04-09
PLANET-3422: Main content in Campaign pages is attached to the bottom of the browser.

pablocubico:
Ref: https://jira.greenpeace.org/browse/PLANET-3422


comzeradd:
Does it make more sense to actually remove this from the [initial mixin](https://github.com/greenpeace/planet4-plugin-blocks/blob/release/v1.40.0/assets/scss/base/_blocks.scss#L5)?

pablocubico:
@comzeradd Very right you are, sir! I also noticed the `_blocks` file contained only a `mixins` import (which is already imported in `style.css`) and it has only a `page-header-block`, so I moved the mixin to the `mixins` file, which seems more obvious and already has a `page-section-header`, and removed the additional `_blocks` file, that doesn't actually hold any "block".

PR #536, opened 2019-04-12
PLANET-3431 restructure commands code

angtheod:
Create class for adding p4-blocks sub-commands.
Moved commands code from main plugin file into this class.
Now we can create sub-commands by adding a method in this class.

### DIFF
--- a/classes/class-loader.php
+++ b/classes/class-loader.php
@@ -8,6 +8,8 @@
 
 namespace P4BKS;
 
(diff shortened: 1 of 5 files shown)


angtheod:
***Note*** WP cli command for replacing shortcode changes from
`wp replace-shortcodes {id}` to `wp p4-blocks replace_shortcodes {id}`

PR #541, opened 2019-04-18
PLANET-3467 Submenu block: problems with styles

dpivo:
Submenu Block Styles Fixed.

Most of the fix involved reverting back to older code.


comzeradd:
@dpivo looks good locally. Can you amend the commit message to be more descriptive? The message should describe what change it made, rather than what problem it solves. See [this guide](https://planet4.greenpeace.org/handbook/dev-git-best-practices/).

Also, it seems that you are fixing two different bugs with this PR. So maybe you can split this to 2 commits.

PR #546, opened 2019-04-25
PLANET-3508 - Remove button class from Split Two Column CTA

comzeradd:
This reverts commit 384514aeca2ed06fb3aae9c517ae90de91baa203.

Not sure why this change happened initially. @pablocubico maybe you have some insight?

### DIFF
--- a/includes/blocks/split_two_columns.twig
+++ b/includes/blocks/split_two_columns.twig
@@ -32,7 +32,7 @@
 					<p class="split-two-column-item-subtitle">{{ issue.description|truncate(25, true)|raw }}</p>
 				{% endif %}
 				{% if ( issue.link_text and issue.link_url ) %}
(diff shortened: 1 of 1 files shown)


pablocubico:
@comzeradd Don't remember **exactly**, but according to: https://jira.greenpeace.org/browse/PLANET-3398, I think I was trying to make sure the Campaign overrides applied to the Split two column block. Can you check if this change collides with the original purpose of that ticket? I guess it should only affect the color. Or maybe ensure it only applies to campaigns and not the whole site.

comzeradd:
It seems safe to remove that class. All campaign themes extend this link to use the `.btn-primary` class. Example: [antarctic theme](https://github.com/greenpeace/planet4-plugin-blocks/blob/develop/assets/scss/campaigns/themes/_theme_antarctic.scss#L146-L148).

PR #547, opened 2019-05-02
PLANET-2849 replace composers autoload

angtheod:
Replace Composer's autoload mechanism with our own, so that the plugin can be used without it.


planet4-blocks.php
```diff
@@ -86,7 +86,7 @@
 	define( 'WP_UNINSTALL_PLUGIN', P4BKS_PLUGIN_BASENAME );
 }
 
-require_once __DIR__ . '/vendor/autoload.php';
+require_once( __DIR__ . '/classes/class-loader.php' );
```
comzeradd: You can skip the parentheses here.

PR #551, opened 2019-05-09
PLANET-3524 - Use consistent vertical spacing on all blocks

comzeradd:
Ref: https://jira.greenpeace.org/browse/PLANET-3524

Removed all custom margins and paddings from blocks. Instead I created relevant classes to be inherited by all blocks.

Changed markup on some blocks to use the more semantic markup tag `<section>`.


comzeradd:
@pablocubico thanks, good catch. I'll remove that, before I merge.

PR #554, opened 2019-05-16
PLANET-3541 - Overrides twig templates if present on child theme

nicklem:


koyan:
@pablocubico Instructions added to ReadMe. I will also add them to the handbook.

PR #556, opened 2019-05-20
Fix checking of requirements for plugin dependencies

oekeur:
Refactors conditional so it is able to return something else than 'false'
This makes sure the hook_plugin() will run


angtheod:
Indeed, this was missed looks correct now, but I would first check the usefulness of the `hook_plugin()` method before merging.

PR #562, opened 2019-05-22
Planet-3565 Counter block - Use API if API is given

koyan:
[Jira issue: PLANET-3565](https://jira.greenpeace.org/browse/PLANET-3565)
Added functionality that allows editors to define an API URL from the global counter application. 
If such a url is added, AND it returns a valid Json, with a field named "unique_count" AND the value is an integer, then the manual "completed" field is overwritten by the one returned by the API. 

To test it: Create a new counter block, and add the url that is defined in the description of the Jira ticket.

### DIFF
--- a/classes/controller/blocks/class-counter-controller.php
+++ b/classes/controller/blocks/class-counter-controller.php
@@ -56,6 +56,15 @@ if ( ! class_exists( 'Counter_Controller' ) ) {
 					],
 				],
 				[
+					'label' => __( 'Completed API URL', 'planet4-blocks-backend' ),
+					'attr'  => 'completed_api',
+					'type'  => 'url',
+					'meta'  => [
+						'placeholder' => __( 'API URL of completed number. If filled in will overide the "Completed" field', 'planet4-blocks-backend' ),
+						'data-plugin' => 'planet4-blocks',
+					],
+				],
+				[
 					'label' => __( 'Target', 'planet4-blocks-backend' ),
 					'attr'  => 'target',
 					'type'  => 'number',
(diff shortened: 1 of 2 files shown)


classes/controller/blocks/class-counter-controller.php
```diff
@@ -98,11 +107,24 @@ public function prepare_fields() {
...
-			$fields['percent'] = $target > 0 ? round( $completed / $target * 100 ) : 0;
-			$fields['text']    = str_replace(
+			$completed = 0;
+			if ( array_key_exists( 'completed', $fields ) ) {
+				$completed = floatval( $fields['completed'] );
+			}
+			$target = floatval( $fields['target'] );
+
+			if ( array_key_exists( 'completed_api', $fields ) ) {
+				$response_api  = wp_remote_get( $fields['completed_api'] );
```
sagarsdeshmukh: We could use `wp_safe_remote_get()` function.

sagarsdeshmukh: We need to catch exception here, if the API endpoint goes down. the page should not break.

koyan: The page does not break. You can test it by putting a wrong url, or an API that returns a different JSON

koyan: > We could use `wp_safe_remote_get()` function.

Done.

koyan: New PR: https://github.com/greenpeace/planet4-plugin-blocks/pull/562

PR #565, opened 2019-05-23
PLANET-3567 - Add class names for tags to article list items

stduerre:
Enables css styling of articles depending on the article's tags.

Opted to put the code on one line to prevent additional space characters in the output.


stduerre:
The implementation has a flaw. I'm going to correct it first and open a PR again.

PR #566, opened 2019-05-24
Feature/planet 3567

stduerre:
Added classes to the article-item-list to be able to style them according to tags in the article


includes/teasers/tease-articles.twig
```diff
@@ -1,4 +1,4 @@
-<article class="article-list-item" >
+<article class="article-list-item{% for tag in recent_post.tags %} tag-{{ tag.slug|e('wp_kses_post') }}{% endfor %}" >
```
comzeradd: Minor tweak: can you move the space before the `{% for %}` loop. I think it would make the code look more readable. And please remove the space at the end of it.

stduerre: It's changed and the commits are squashed now.
The reason I originally put the space inside the loop was because that way the output doesn't contain a space when there's no additional class to add. But it won't affect functionality and I agree the code is more readable like this.

PR #567, opened 2019-05-24
PLANET-3557 Counter Block: Implement design changes

dpivo:
- Design changes implement on the three different counter block options


assets/scss/components/_counter.scss
```diff
@@ -47,5 +53,10 @@
 
   .counter-text {
     margin-bottom: 0;
+    font-size: 1rem;
+  }
+
+  .counter-text-arc {
+    font-weight: bold;
```
comzeradd: I think the intention of the design here was for that text to be bold when the goal is reached, regardless if it's arc or progress bar. Can you confirm with Will?

PR #568, opened 2019-05-24
PLANET-3566 Add to the plugins report  which tags are use redirection

koyan:
[JIRA ticket](https://jira.greenpeace.org/browse/PLANET-3566)
Add to the plugins report with a breakdown which tags are use redirection and which do not. 
Add it to both the report in the backend, and the json exported. 
While this funcionality (tags with redirection) is not part of the plugins block but of the master theme, we add it to the plugins report, because it is already exporting data to the external place we want to report to.

Note: After all NROs/sites have switched all their tag pages to redirect instead of auto generation, we will retire the functionality for redirection, and we will then retire this part of the report as well.

You can demo the functionality 
a) By seeing the report in the Backend -> Blocks -> usage menu
3566-tags-redirection

b) by going to the url /wp-json/plugin_blocks/v1/plugin_blocks_report/  on your site and checking the `TagsUsingRedirectionPage` and `TagsNotUsingRedirectionPage` values

### DIFF
--- a/classes/controller/menu/class-blocks-usage-controller.php
+++ b/classes/controller/menu/class-blocks-usage-controller.php
@@ -115,6 +115,42 @@ if ( ! class_exists( 'Blocks_Usage_Controller' ) ) {
 				echo '<br>';
 			}
 
+
+			// Add to the report a breakdown of which tags are using a redirect page and which do not
+			// The first query shows the ones that do not use a redirect page
+			$sql     = "SELECT term.name, tm.meta_value, tt.term_id
+						FROM " . $wpdb->prefix . "term_taxonomy AS tt, " . $wpdb->prefix . "terms AS term, " . $wpdb->prefix . "termmeta AS tm
+						WHERE tt.`taxonomy`='post_tag' 
+						AND term.term_id = tt.term_id
+						AND tm.term_id=tt.term_id
+						AND tm.meta_key='redirect_page'
+						AND tm.meta_value =''";
+			$results = $wpdb->get_results( $sql );
+			echo '<hr>';
+			echo '<h2>Tags without redirection page</h2>';
+			foreach ( $results as $result ) {
+				echo '<a href="term.php?taxonomy=post_tag&tag_ID=' . $result->term_id . '" >' . $result->name . '</a>';
+				echo '<br>';
+			}
+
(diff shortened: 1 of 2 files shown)


classes/controller/menu/class-blocks-usage-controller.php
```diff
@@ -115,6 +115,42 @@ public function plugin_blocks_report() {
...
+			$sql     = "SELECT term.name, tm.meta_value, tt.term_id
+						FROM " . $wpdb->prefix . "term_taxonomy AS tt, " . $wpdb->prefix . "terms AS term, " . $wpdb->prefix . "termmeta AS tm
+						WHERE tt.`taxonomy`='post_tag' 
+						AND term.term_id = tt.term_id
+						AND tm.term_id=tt.term_id
+						AND tm.meta_key='redirect_page'
+						AND tm.meta_value =''";
+			$results = $wpdb->get_results( $sql );
+			echo '<hr>';
+			echo '<h2>Tags without redirection page</h2>';
```
sagarsdeshmukh: please use translation function `__('text', 'planet4-blocks-backend')`.

classes/controller/menu/class-blocks-usage-controller.php
```diff
@@ -115,6 +115,42 @@ public function plugin_blocks_report() {
...
+			$sql     = "SELECT term.name, tm.meta_value, tt.term_id
+						FROM " . $wpdb->prefix . "term_taxonomy AS tt, " . $wpdb->prefix . "terms AS term, " . $wpdb->prefix . "termmeta AS tm
+						WHERE tt.`taxonomy`='post_tag' 
+						AND term.term_id = tt.term_id
+						AND tm.term_id=tt.term_id
+						AND tm.meta_key='redirect_page'
+						AND tm.meta_value !=''";
+			$results = $wpdb->get_results( $sql );
+			echo '<hr>';
+			echo '<h2>Tags that use a redirection page</h2>';
```
sagarsdeshmukh: same here.

sagarsdeshmukh:
Except string translation, PR looks good to me 👍

koyan:
@sagarsdeshmukh We are not going to translate that. This is only for the usage of our community manager to communicate for now with the NROs which pages they need to modify.

PR #592, opened 2019-06-28
PLANET-3634 Columns: align headers vertically

comzeradd:
Ref: https://jira.greenpeace.org/browse/PLANET-3634

So this one was a bit tricky. The Columns block layout changes a lot between breakpoints (eg. we stack it completely on mobile). So there is no easy way to align elements between different columns.

On tasks variation we have converted the [markup](https://github.com/greenpeace/planet4-plugin-blocks/blob/develop/includes/blocks/columns_tasks.twig) to achieve horizontal grid, but in general it's a bad practice (a11y-wise) to adjust markup to presentation. Screen readers will read that horizontally :/ So we should probably fix that one too.

I also tried actual css-grid, but it's hard to make things float properly in a columns direction. Especially when things keep change between breakpoints.

I ended up with this js hack. It's still not perfect. For instance, we probably only want to adjust height on header that on the same row. Not sure yet how to achieve that.

@pablocubico feedback/ideas?


comzeradd:
After getting some feedback from Will, I'm merging this. I added an extra check, so at least on mobile we don't mess with the headers.

PR #593, opened 2019-07-02
PLANET-3635 Make the Progress Bar match the Call To Action color when the counter is inside the form block

dpivo:
- Added a condition to change the colour of the progress bar to the colour of the take action button if it is inside an en-form


includes/blocks/counter.twig
```diff
@@ -14,7 +14,7 @@
             <div class="content-counter">
                 {% if fields.style == 'bar' or fields.style == 'en-forms-bar' %}
                     <div class="progress-container">
-                        <div class="progress-bar" style="width: calc( {{ fields.percent }}% + 20px);"></div>
+                        <div class="progress-bar {{ 'en-forms-bar' == fields.style ? 'enform-progress-bar' }}" style="width: calc( {{ fields.percent }}% + 20px);"></div>
```
angtheod: Shouldn't it be? `{{ 'en-forms-bar' == fields.style ? 'enform-progress-bar' : '' }}`

pablocubico: I just googled it, looks like you can omit the falsey argument with Twig: https://twig.symfony.com/doc/2.x/templates.html#other-operators

PR #606, opened 2019-07-17
PLANET-3638 - JS: Use the same closure method for all js code

pablocubico:
Ref: https://jira.greenpeace.org/browse/PLANET-3638


27Bslash6:
See: https://github.com/greenpeace/planet4-master-theme/pull/843

PR #608, opened 2019-07-19
PLANET-3740 CPP - Disable layout versions of Covers block in Campaign content type

dpivo:
- first two styles of covers disabled if post type is campaign
- remaining cover option automatically selected


includes/blocks/covers.twig
```diff
@@ -12,6 +12,8 @@
 
 		<section class="block covers-block {{ covers_view_class }}">
 			<div class="container">
+				pooooop
```
kirdia: please remove the pooo :)

classes/controller/blocks/class-newcovers-controller.php
```diff
@@ -51,31 +51,8 @@ public function print_admin_footer_scripts() {
...
-						],
-						[
-							'value' => '3',
-							'label' => __( 'Content Covers', 'planet4-blocks-backend' ),
-							'desc'  => 'Content covers pull the image from the post.',
-							'image' => esc_url( plugins_url() . '/planet4-plugin-blocks/admin/images/content_covers.png' ),
-						],
-					],
-				],
+
```
kirdia: @dpivo you could populate the options in a different array and pass it to the field options. this way will reduce duplication

PR #615, opened 2019-07-25
PLANET-3638 - Use the same closure for DOM ready events.

pablocubico:
See: https://github.com/greenpeace/planet4-master-theme/pull/843


pablocubico:
@27Bslash6 this new PR changes the `$(document).ready...` calls into `jQuery(function ($) { ...`, which is the recommended way according to the deprecation docs, not an IIFE, but the `$` is now received as an argument for the callback.

admin/js/cookies.js
```diff
@@ -10,7 +10,7 @@ function setNoTrackCookie()
   }
 }
 
-$(document).ready(function() {
+jQuery(function ($) {
```
27Bslash6: The function setNoTrackCookie is still vulnerable to $ pollution here.  Perhaps move into the document ready handler?  Is it used inline anywhere else?

27Bslash6:
Are these test failures related?

admin/js/cookies.js
pablocubico: Good catch! Looks like it's not, I'll move it to the ready handler. I noticed a few other contexts accessing $ globally, as the code for the blocks in the admin side, which could probably be enclosed better, but I won't engage in refactoring it since it's going to be deprecated soon in favor of the Gutenberg versions. The JS for the "frontend" site will remain mostly untouched (ideally).

kirdia:
> Are these test failures related?

Nope, it's related to an engaging networks related issue.

comzeradd:
Similarly to the other PR, `style.css.map` is modified (thus the conflict), but there is no css change introduced here.

PR #619, opened 2019-07-30
PLANET-3840 Column block: display block when no column header is filled in

dpivo:
- Changed to allow the block to display even if the column header is empty


classes/controller/blocks/class-columns-controller.php
```diff
@@ -208,19 +202,27 @@ public function prepare_data( $attributes, $content = '', $shortcode_tag = 'shor
...
-				$attributes_temp = array_merge( $attributes_temp, $column_atts );
 
-				if ( ! empty( $attributes[ "title_$i" ] ) ) {
-					$columns_set = $i;
+				if ( empty( $column_atts[ "title_$i" ] ) ) {
+					foreach ( $column_atts as $key => $column_att ) {
+						if ( ! empty( $column_att ) && "link_new_tab_$i" !== $key ) {
+							$columns_set = $i;
+
+							$column_atts[ "title_$i" ] = ' ';
```
kirdia: @dpivo why are you setting the title here? it will render an empty heading, is this needed?
maybe you could change the twig template if we need to render the column, despite the title being empty.

dpivo: Changed the logic

angtheod:
I think PLANET-3825 (#620) and PLANET-3840 (#619) should be one ticket because the changes are in the same code lines.
Not sure if this PR does something extra but after talking with Will, I removed the header field from being compulsory and added the description, attachment fields instead without any JS changes.

These 2 PRs are conflicting each other. What should we do?
@kirdia @sagarsdeshmukh @dpivo

dpivo:
> I think PLANET-3825 (#620) and PLANET-3840 (#619) should be one ticket because the changes are in the same code lines.
> Not sure if this PR does something extra but after talking with Will, I removed the header field from being compulsory and added the description, attachment fields instead without any JS changes.
> 
> These 2 PRs are conflicting each other. What should we do?
> @kirdia @sagarsdeshmukh @dpivo

@angtheod I think we have different logic. the change I just added allows any of the fields to be filled in for the column to appear. It looks like you are changing the mandatory field from the the title to the description. So I think that basically if your change is fine and they want those fields to be mandatory then only your change is necessary and we can scrap mine

angtheod:
> @angtheod I think we have different logic. the change I just added allows any of the fields to be filled in for the column to appear. It looks like you are changing the mandatory field from the the title to the description. So I think that basically if your change is fine and they want those fields to be mandatory then only your change is necessary and we can scrap mine

Ok, I confirmed yesterday with Design. Body (description) and Image (attachment) will be the new mandatory fields for the block to appear. So, I close this PR for now, ok?

PR #622, opened 2019-08-05
PLANET-3793: Blocks in Posts: Timeline

pablocubico:
Ref: https://jira.greenpeace.org/browse/PLANET-3793


assets/scss/components/_timeline.scss
```diff
@@ -193,3 +193,127 @@
...
+      font-size: 1.5rem;
+      line-height: 1.87rem;
+      margin-bottom: $n20;
+      padding: 0 !important;
+    }
+
+    .tl-timeline {
+      .tl-slidenav-next {
+        display: none;
+        @include small-and-up {
```
comzeradd: nit: add a line above the `@include`, here and a few lines below.

PR #623, opened 2019-08-05
PLANET-3819 CPP - In-block styling for 'Full Width' Carousel block.

dpivo:
- Header size option added for full width classic carousel block


comzeradd:
This change seems to be breaking the two CarouselHeader tests (check CI output). Can you open a PR to base-fork to amend these tests?

PR #625, opened 2019-08-07
PLANET-3783 Gallery Block: Grid & 3 Column styles are not consistent with the Slider style

dpivo:
- left and right padding removed from the container div on the 3 column and grid


comzeradd:
I think we are going to meet this problem on other blocks in posts too, since we use `container` class on all of them.

I think it's better to add something like this, so it's applied to all containers inside post content:
```
div.post-content {
  .container {
    padding-left: 0;
    padding-right: 0;
  }
}
```

This can live on master-theme on the [_posts partial](https://github.com/greenpeace/planet4-master-theme/blob/develop/assets/scss/pages/post/_post.scss).

We have something similar for pages on [_page partial](https://github.com/greenpeace/planet4-master-theme/blob/develop/assets/scss/pages/_page.scss#L21-L24).

dpivo:
> I think we are going to meet this problem on other blocks in posts too, since we use `container` class on all of them.
> 
> I think it's better to add something like this, so it's applied to all containers inside post content:
> 
> ```
> div.post-content {
>   .container {
>     padding-left: 0;
>     padding-right: 0;
>   }
> }
> ```
> 
> This can live on master-theme on the [_posts partial](https://github.com/greenpeace/planet4-master-theme/blob/develop/assets/scss/pages/post/_post.scss).
> 
> We have something similar for pages on [_page partial](https://github.com/greenpeace/planet4-master-theme/blob/develop/assets/scss/pages/_page.scss#L21-L24).

Done, I'll close this PR because there will be one against master theme.

PR #627, opened 2019-08-08
PLANET-3902 columns new specs

angtheod:
Apply new specs for displaying columns in Columns block.
Header or Body need to be filled in order for a column to appear.

### DIFF
--- a/includes/blocks/columns.twig
+++ b/includes/blocks/columns.twig
@@ -18,12 +18,13 @@
 
 					<div class="row">
 						{% for i in 1..4 %}
-							{% if fields['title_'~i] %}
-								{% set title = fields['title_'~i] %}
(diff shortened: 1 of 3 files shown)


includes/blocks/columns.twig
```diff
@@ -38,9 +39,7 @@
 											{{ title|e('wp_kses_post')|raw }}
 										{% if link and not is_campaign %}</a>{% endif %}
 									</h3>
-									{% if description %}
-										<p>{{ description|e('wp_kses_post')|raw }}</p>
-									{% endif %}
+									<p>{{ description|e('wp_kses_post')|raw }}</p>
```
comzeradd: I think we still need the check here to avoid having an empty `<p>` element when the description is empty.

angtheod: fixed

PR #628, opened 2019-08-09
PLANET-3896 - Carousel Header: Hide action-button link instead of div

comzeradd:
Ref: https://jira.greenpeace.org/browse/PLANET-3896

We use the action button as a point of reference, but on Campaigns we hide it. This was breaking parts of the Carousel functionality.

@pablocubico I'm adding you only as reviewer, since you have more experience with Carousel's js code.


pablocubico:
Hey @comzeradd, as this is used to align the indicators, it looks like it's breaking at some sizes because of the padding in `main-header`.  How about using the description or the `h1` on top? I think the title is required. 🤔 

image

comzeradd:
@pablo I thought of that, but we have two problems:
1. h1 can also be h2 or h3
2. Besides offset, we also use width to align the indicators. And header has a completely different width behavior than action-button.

`.action-button` is the class we use on the button container, not the actual button. So I amended the code, in order to hide the button link instead of the div. It's not perfect to render an empty div, but it's simpler code-wise. When I started playing with padding and {inner|outer}Width the code started to get really unreadable.

PR #633, opened 2019-08-14
PLANET-3794-vs-fix fixed vertical spacing on task column in posts

dpivo:
- Small change to vertical spacing of task numbers for tasks in posts


assets/scss/components/_columns.scss
```diff
@@ -723,13 +723,15 @@
             height: 100px;
             font-size: 3rem;
             line-height: 90px;
+            margin-bottom: 32px;
```
comzeradd: You can use our standard [spacing variables](https://github.com/greenpeace/planet4-styleguide/blob/master/src/base/_variables.scss#L40) for those values

PR #635, opened 2019-08-21
PLANET-3838 Customise Take Action boxout block to allow custom URLs

dpivo:
- Added custom fields to Take Action Boxout


comzeradd:
@dpivo the "Open in a new Tab" seems to remain active, after I choose something from the dropdown.

PR #647, opened 2019-09-03
PLANET-38383-edit-fix fixed edit screen being blank

dpivo:
- Small change to fix checking undefined variable

PR #649, opened 2019-09-05
PLANET-3838 Customise Take Action boxout block to allow custom URLs

dpivo:
- Add custom fields to allow customisation of Take Action Boxout


comzeradd:
we should close #647 right?

PR #650, opened 2019-09-06
Fix stylelint and eslint paths

comzeradd:
Double asterisk patterns need double quotes to work properly.


kirdia:
can you add it for eslint also?

comzeradd:
@kirdia done!

PR #664, opened 2019-09-26
PLANET-4082 CPP - Incorrect and inconsistent counter numbers

dpivo:
- added transient variable to keep counter value consistent.

### DIFF
--- a/classes/controller/blocks/class-counter-controller.php
+++ b/classes/controller/blocks/class-counter-controller.php
@@ -138,7 +138,15 @@ if ( ! class_exists( 'Counter_Controller' ) ) {
 			$target    = isset( $fields['target'] ) ? floatval( $fields['target'] ) : 0;
 
 			if ( array_key_exists( 'completed_api', $fields ) ) {
-				$response_api = wp_safe_remote_get( $fields['completed_api'] );
+				if ( get_transient( $fields['completed_api'] ) ) {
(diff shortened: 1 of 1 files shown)


classes/controller/blocks/class-counter-controller.php
```diff
@@ -138,7 +138,13 @@ public function prepare_data( $fields, $content, $shortcode_tag ) : array {
 			$target    = isset( $fields['target'] ) ? floatval( $fields['target'] ) : 0;
 
 			if ( array_key_exists( 'completed_api', $fields ) ) {
-				$response_api = wp_safe_remote_get( $fields['completed_api'] );
+				if ( get_transient( $fields['completed_api'] ) ) {
+					$response_api = get_transient( $fields['completed_api'] );
+				} else {
+					$response_api = wp_safe_remote_get( $fields['completed_api'] );
+					$one_hour     = 3600;
+					set_transient( $fields['completed_api'], $response_api, $one_hour );
```
kirdia: I think it's better not to store the transient the result from the http request is a WP_Error, it could lead to inconsistent data.

kirdia: yes, check if it returns a WP_Error instance and store it if it returns a response

kirdia:
looks good, there are some styling errors though

